PR #6, opened 2015-07-13
Testing

danni:


.jscsrc
```diff
@@ -1,13 +1,17 @@
 {
     "preset": "airbnb",
+
     "fileExtensions": [".js", ".jsx"],
     "excludeFiles": ["node_modules/**"],
+
+    "esprima": "./node_modules/babel-jscs",
```
DanielHeath, line 7: What does `esprima` refer to here?

danni, line 7: It's the JS parser that jscs uses.

danni, line 7: http://esprima.org/

PR #15, opened 2015-07-15
Docker

danni:


Makefile
```diff
@@ -0,0 +1,77 @@
+APP = ask_izzy
+REPO = contyard.office.infoxchange.net.au/$(APP)
```
DanielHeath, line 2: Is that where the source is hosted internally?

Makefile
```diff
@@ -0,0 +1,77 @@
+APP = ask_izzy
```
DanielHeath, line 1: Should we use the Makefile in place of a directory full of scripts?

Makefile
danni, line 2: That's the internal Docker container yard that Straddle Carrier uses to load containers.

Makefile
danni, line 1: Can do? Flexible.

PR #16, opened 2015-07-15
WIP: Attempt to use material-ui with fluxible

DanielHeath:
Current status: themes are not passed through to components resulting in render errors


DanielHeath:
This is unlikely to end up working out. 

As discussed on Monday, I'm going to try removing flux (we don't need it in a read-only app).

PR #17, opened 2015-07-19
Patches to docker config

DanielHeath:


.githooks/pre-commit
```diff
@@ -39,10 +39,15 @@ EOF
...
+READ_TWO_CHARS="-N2"
+if uname -a | grep Darwin > /dev/null ; then
+    READ_TWO_CHARS="-n2"
+fi
+
 # Check for bogusly executable files
 for file in $(git ls-files); do
     if test -x "$file"; then
-        read -N2 hashbash < $file
+        read "$READ_TWO_CHARS" hashbash < $file
```
danni, line 50: This might be more consistent if we use 'bash' or 'dash' instead of trusting whatever 'sh' is.

invoke.sh
```diff
@@ -8,10 +8,8 @@
 APP="s2x_project"
 APP_USER="app"
 
-INVOKE=$0
-
-: ${WEB_CONCURRENCY:=1}
-: ${WORKER_CONCURRENCY:=1}
+: "${WEB_CONCURRENCY:=1}"
+: "${WORKER_CONCURRENCY:=1}"
```
danni, line 12: This looks wrong.

PR #18, opened 2015-07-21
Make release-staging work

danni:


brettrann:
That's, like, readable now!

++

PR #22, opened 2015-07-22
Replace fluxible with react-router

DanielHeath:
At this point, the isomorphic stuff works beautifully and the app is drastically simpler.


src/server/render.js
```diff
@@ -2,80 +2,47 @@
...
+            const html = React.renderToStaticMarkup(
+              <HtmlDocument
+                markup={markup}
+                script={webpackStats.script}
+                css={webpackStats.css}
+              />
+            );
+            const doctype = "<!DOCTYPE html>";
+            res.send(doctype + html);
+        });
```
danni, line 40: I don't understand why we're running the router in 2 places.

DanielHeath, line 40: Once on the server (when a request comes in) and once on the client (on `window.history.push`)

src/routes.js
```diff
@@ -1,34 +1,26 @@
...
-    },
-
-    // This route doesn't point to any handler.
-    bad: {
-        path: "/bad",
-        method: "get",
-        action: InitActions.badPage,
-    },
-};
+export default <Router.Route
```
danni, line 10: Syntax error in Travis. Should this be a .jsx file? Do you need to wrap the XML node in brackets?

DanielHeath, line 10: just looking at this now.

PR #23, opened 2015-07-23
Rename 'assets' to 'static'

DanielHeath:


danni:
In the prod server we don't want to serve `/static/` -- although we will for tests, so I think we need to create an intermediate env for the tests that runs with all debug flags off but still serves `/static`.

In `invoke.sh`:`deploy` we need to copy the assets to `/static`

danni:
Tests run with the prod config, but need assets.

PR #24, opened 2015-07-23
Retry npm install when building on travis - it fails often

DanielHeath:


danni:
Add a comment explaining why.

danni:
Let's replace this by Travis caching if it's available for containerised builds.

PR #29, opened 2015-07-27
Screenshot each component so you can visually compare revisions of them.

DanielHeath:


DanielHeath:
So, according to saucelabs, Firefox on Linux can't do icon fonts :-1:

DanielHeath:
Looks ok on chrome though: https://github.com/ask-izzy/ask-izzy/blob/4fdea1b5314a734e0472ee97de10ed2598fc002f/src/components/CategoryListItem/chrome-43.0.2357.65%20985x739.png

gibbage:
Which Linux? We use icon fonts all across realestate.com.au and it looks fine on my Ubuntu computer at home.

DanielHeath:
Saucelabs linux/firefox. They have a blog post about it but it didn't really help me understand what's wrong with their config.

danni:
Do we want more than one set of reference screenshots?

DanielHeath:
Chrome & Firefox tend to be very similar. I think there's value in one set per tested platform if we decide to target mobile and IE.

danni:
Also failing tests for not being pixel perfect is probably not a useful thing IMO.

DanielHeath:
I've separated the behavior tests from the visual tests so that the 'main' tests don't fail whenever anything changes. 

The more useful outcome of the visual tests is that each PR ends up including 'how does this change the look' inline.

PR #36, opened 2015-08-04
Add ISS service

DanielHeath:
@danni I'm still getting 401 from the browser, even with the staging keys. 

I think it's CORS preventing the basic auth header being set - do you remember any issue like that?


danni:
No issues except for tests failing.

PR #38, opened 2015-08-05
Cleanup

DanielHeath:


.githooks/pre-commit
```diff
@@ -57,4 +57,6 @@ done
 # If there are whitespace errors, print the offending file names and fail.
 git diff-index --check --cached $against --
 
+npm shrinkwrap --dev
```
danni: Bad idea IMO.

src/components/Chevron.js
```diff
@@ -1,19 +1,12 @@
...
-            <i
-                className="material-icons"
-                style={{
-                    marginLeft: "auto",
-                }}
-            >
-                {/* icon: chevron_right */}
-                &#xE5CC;
-            </i>
+            <Icon iconCode="&#xE5CC;" iconName="chevron_right" />
```
danni: Can we move the codes into a hash?

.githooks/pre-commit
DanielHeath: Could you elaborate? 

I usually prefer to know which version of my dependencies I'm going to run.

src/components/Chevron.js
DanielHeath: Good idea.

.githooks/pre-commit
danni: You should update the deps manually. Not as part of the commit hook. Otherwise it can lead to surprises and/or insanity.

DanielHeath: Shrinkwrap doesn't update versions of things - it records what you currently have installed, and fails if you have extra stuff installed or missing packages.

DanielHeath: More generally, I'd like to fail the commit if you add a new dependency and don't shrinkwrap it. Not actually fussed as to how we do that.

danni: Oh, what writes out the JSON file then? I might be confused by the tools. I'm more than happy for something that fails if you haven't updated package.json.

danni:
++

PR #39, opened 2015-08-06
Check dependencies are installed correctly before committing.

DanielHeath:


DanielHeath:
Splitting this out of https://github.com/ask-izzy/ask-izzy/pull/38 to avoid blocking that PR.

danni:
Having it update it automatically is not ideal IMO.

I'd be happier with a test that says "this doesn't match and that is bad". Since it might be your env that is wrong i.e. if you update your env, and update the shrinkwrap and then I hack some code without updating my env and then just write over it again.

It's be better if something went "your current env doesn't match the shrinkwrapping, update your env or the shrinkwrapping" and that caused the pre-commit to fail.

DanielHeath:
So, while there are several implementations of that on NPM, none of them actually caused any errors when I set up an invalid state.

I've written one and tested that it fails with a useful message on all permutations of extra/missing dependencies.

danni:
++

PR #42, opened 2015-08-06
Visual tests run each platform in travis.yml

DanielHeath:


danni:
Consider breaking BROWSERS out into a separate file?

DanielHeath:
Will travis read a separate file?

danni:
Don't know TBH. Yaml can use named nodes, so maybe?

DanielHeath:
I'd be suspicious of a YAML implementation that allowed parsing other files (can I pull in `/etc/passwd`)?

DanielHeath:
I don't see any support for something like that in https://github.com/travis-ci/travis-yaml/blob/master/lib/travis/yaml/parser/psych.rb

DanielHeath:
Any other concerns?

danni:
Okay then. ++

PR #44, opened 2015-08-09
Style category list heading according to design brief

DanielHeath:


DanielHeath:
Lets do https://github.com/ask-izzy/ask-izzy/pull/45 first

PR #46, opened 2015-08-09
Visual snapshots

DanielHeath:


danni:
Why are we still getting random XML entities in our screenshots?

DanielHeath:
Not just the screenshots - I think webpack is double-escaping something in JSX. 
If you use the server-side render it looks fine on more platforms; the client-side JS doesn't and I haven't looked yet, since we're about to rip out all those icons in favor of the navy design ones which might not have the same issue.

danni:
Let's put in the Navy ones and see if we still have the same problem.

DanielHeath:
Good to merge these in the meantime?

danni:
Also lots of these screenshots still have broken XML attrs in them.

PR #48, opened 2015-08-11
A few little fixes

DanielHeath:


DanielHeath:
I should actually split this out into multiple PRs

PR #53, opened 2015-08-13
Fixup: check-shrinkwrap should ignore all dotfiles (eg .DS_Store

DanielHeath:


danni:
Is .DS_Store not in your global .gitignore ?

danni:
Or does shrinkwrap not consider that. Should it do that instead?

DanielHeath:
`node_modules` is gitignored - I'm just comparing directory listings to see if you've got deps installed

danni:
++

PR #56, opened 2015-08-18
Add script to automatically convert svg icons to react components

DanielHeath:


danni:
Let's not add Ruby to our dep chain.

DanielHeath:
I'm sure porting would be straightforward enough at some point; can we merge the icons without the script? Want to keep moving & not spend time figuring out XML libs in node.

PR #59, opened 2015-08-18
Changed to updated sky blue palette

mucow:
Updated the colours. It's failing a new recent test for categories though.


DanielHeath:
++

PR #60, opened 2015-08-19
Cleaning up a few small bits and pieces

DanielHeath:


src/constants/categories.js
```diff
@@ -18,8 +18,8 @@ class Category {
         this.byline = props.byline;
         this.icon = props.icon;
         this.key = this.name
-            .replace(' ', '-')
-            .replace('&', '-and-')
+            .replace(/&/g, 'and')
+            .replace(/ /g, '-')
```
danni: For slugs, consider something like `/[^A-Za-z0-9-_]/`

src/server.js
```diff
@@ -7,6 +7,11 @@ import render from "./server/render";
 
 // Initialize express server
 
+if (!process.env.ISS_URL) {
+    console.log("Must set ISS_URL");
```
danni: `console.error`

PR #62, opened 2015-08-20
Style service details

DanielHeath:


src/components/Phone.js
```diff
@@ -0,0 +1,29 @@
...
+
+    number(): string {
+        if (this.props) {
+            return this.props.number;
+        }
+        return ""
+    }
+
+    href(): string {
+        return "tel:" + this.number().replace(/[^0-9]/g, '');
```
danni: Add `+` to the list of allowed chars just in case.

src/pages/ServicePage.js
```diff
@@ -26,28 +26,84 @@ class CategoryPage extends React.Component {
...
+                </mui.Paper>
+                <mui.Paper>
+                    <OpeningTimes object={object} />
+                </mui.Paper>
+                <mui.Paper>
+                    <Address {...object.location} />
+                </mui.Paper>
+                {this.renderPhones(object.phones)}
+                <mui.Paper>
+                    <h3>To receive this service you must be</h3>
```
danni: Be aware these may not be set. In which case I don't think we should show that particular heading.

src/pages/ServicePage.js
```diff
@@ -26,28 +26,84 @@ class CategoryPage extends React.Component {
...
                 this.setState({
                     object: data,
                 });
             });
 
-        // FIXME: display error
+        // FIXME: display error on failure to connect
+    }
+
+    renderPhones(phones: Array<Object>): React.Element {
```
danni: We probably want to do something useful with the different phone types. Only show some of the phone types etc.

Phone types are `phone`, `mobile`, `fax`, `tty` and `freecall`.

src/components/Address.js
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        return <span className={property} >{described}</span>;
+    }
+
+    render(): React.Element {
+        return (
+            <div className="Address">
+                {this.describe('level', 'Level')}
+                {' '}
```
danni: `&nbsp`? Since we don't want to break across the address anyway.

src/components/Address.js
```diff
@@ -0,0 +1,43 @@
...
+                {this.props.street_number}
+                {' '}
+                {this.props.street_name}
+                {' '}
+                {this.props.street_type}
+                {' '}
+                {this.props.street_suffix}
+                {' '}
+                {this.props.suburb}
```
danni: Split into a second line here though.

danni:
Merge in master and write a test for it on top of the mock ISS.

src/components/Address.js
DanielHeath: multiple consecutive `&nbsp`s will render if there is no content which isn't ideal...

danni:
Is ServicePane still used?

danni:
++

PR #63, opened 2015-08-20
Tests for results page

danni:
Add tests for the results page and a mock ISS server for functional tests.


src/components/OpeningTimes.js
```diff
@@ -24,11 +24,11 @@ class OpeningTimes extends React.Component {
 
         var todayOpen = object.opening_hours[
             moment().day()
-        ];
+        ] || {};
 
         var nextOpen = object.opening_hours[
             (moment().day() + 1) % object.opening_hours.length
-        ];
+        ] || {};
```
DanielHeath, line 31: This looks like a hack; its OK for now but I'd like to find a better approach.

DanielHeath:
++

src/components/OpeningTimes.js
danni, line 31: Yeah. I'm actually not 100% sure I'm doing this right anyway. It might be we need to use `_.findWhere` to find the right day.

PR #65, opened 2015-08-20
Add script to generate components.js

DanielHeath:


DanielHeath:
Generate rather than hand-maintaining this list

danni:
++ if tests pass.

danni:
++

PR #66, opened 2015-08-20
Share fixtures between mock server and styleguide

DanielHeath:


danni:
++ if tests pass

DanielHeath:
IE build failed due to saucelabs dropping the connection

PR #67, opened 2015-08-20
Refactor header bar

danni:


src/pages/HomePage.js
```diff
@@ -8,16 +7,34 @@ class HomePage extends React.Component {
 
     render(): React.Element {
         return (
-            <div className="Page">
-                <div className="Page-header">
-                    <HeaderBar />
+            <div className="HomePage">
+                <div className="header">
+                    <div className="branding-container">
```
DanielHeath, line 12: Should the brand/logo stuff still be a component?

danni, line 12: It only appears once, it may as well be on the page IMO.

src/pages/CategoryPage.js
```diff
@@ -60,7 +60,6 @@ class CategoryPage extends React.Component {
     render(): React.Element {
         return (
             <div>
-                <Router.RouteHandler />
```
DanielHeath, line 63: Cool - just bear in mind if we nest anything under this URL it won't render.

src/pages/HomePage.js
DanielHeath, line 12: Fair

DanielHeath:
++

src/pages/CategoryPage.js
danni, line 63: It was really inconsistently used. It seemed better to clean it up for now.

PR #68, opened 2015-08-20
Style category page

victornguyen:
Styling work done to CategoryPage components done as part of Hack Day.

Apologies in advance!


src/pages/CategoryPage.js
```diff
@@ -60,6 +61,8 @@ class CategoryPage extends React.Component {
                 <Router.RouteHandler />
                 <mui.AppBar title={this.category.name} />
 
+                <CategoryHeader />
```
danni: In #67 I refactored HeaderBar to use for things like this. Because I think it's consistently the same styles.

src/components/ResultTile.js
```diff
@@ -53,11 +54,15 @@ class ResultTile extends React.Component {
                     />
                 }
             >
-                <div>{object.name}</div>
-                <div>{object.site.name}</div>
+                <h2 className="name">{object.name}</h2>
+                <div className="site">{object.site.name}</div>
```
danni: Can you merge this with master? The class here is site_name, which is being used in a test. Happy to have either, so long as the test is correct.

src/components/ResultTile.js
```diff
@@ -53,11 +54,15 @@ class ResultTile extends React.Component {
...
                 <OpeningTimes object={object} />
-                <div>FIXME transport time</div>
-                <div>{object.service_types[0]}</div>
+                <TransportTime object={object} />
+                <ul className="service-list">{
+                    (object.service_types || []).map((service) => {
+                        return <li>{service}</li>;
+                    })
```
danni: I believe you can do `map(service => <li>{service}</li>)`

src/components/CategoryHeader.js
```diff
@@ -0,0 +1,21 @@
...
+                <div className="found">
+                    I found 25 housing services for men
+                    ages 25-54 in Brunswick
+                </div>
+                <a className="search-link" href="#">
+                    Try a different search
+                </a>
+            </div>
+        );
+    }
```
danni: This probably doesn't need its own class IMO. Also you can get (some of) the live data for this field from `this.state.meta` on the category page.

src/components/OpeningTimes.js
```diff
@@ -35,21 +37,24 @@ class OpeningTimes extends React.Component {
         ] || {};
 
         if (object.now_open.now_open) {
-            // get closing time
             return (
-                <div>
-                    <span>Now open</span>&nbsp;
-                    <span>Closes {formatTime(todayOpen.close)}</span>
+                <div className="OpeningTimes">
+                    <icons.Clock className="coloured-icon" />
```
danni: #71 includes a refactoring of the coloured icons to make it much easier to use.

src/components/CategoryHeader.js
danni: Either include this in `CategoryPage` or use `HeaderBar` as a generic element.

DanielHeath: this.state.meta is just a huge URL string

danni: Oops. Someone should debug that.

PR #72, opened 2015-08-21
Fix a couple of niggling tooling issues

DanielHeath:


script/generate-component-index
```diff
@@ -23,3 +24,22 @@ do
...
+do
+    base=$( basename "$f" )
+    echo "@import '../components/$base';" >> $target
+done
+
+for f in $( find src/pages -name '*.scss' | sort )
+do
+    base=$( basename "$f" )
+    echo "@import '../pages/$base';" >> $target
+done
```
danni, line 39: Rather than repeating, why not just all of `src`? In which case we could call it something instead of `components.scss`.

DanielHeath, line 39: That would include `src/styles`, which has (eg) color variables that have to be included up-front.

danni, line 39: You can exclude a dir.

DanielHeath, line 39: I don't see that option in BSD `find`

danni, line 39: You can give a predicate multiple times and you canprepend with a `!` so

```
! -name ...
```

DanielHeath, line 39: done

danni:
++

PR #73, opened 2015-08-21
Clean up icon usage in ResultTile

danni:


kane-c:
Can `ResultTile.scss` be removed now?
Otherwise ++

PR #74, opened 2015-08-21
Navigation between pages

danni:
Use the router properly so we can navigate between pages without making a request to the server. Also support going back using the router context.


kane-c:
Looks ok to me but might want a second opinion

PR #78, opened 2015-08-24
Style work, router fixes and other bits and pieces

danni:
Builds off #77.

Branch of misc:
- Style work across various pages
- Fix the use of the router in listitems _properly_ (has tests to validate behaviour :-P)
- Other misc like flow types, mock improvements, etc.


DanielHeath:
++

PR #83, opened 2015-08-25
Fix styleguide pages

DanielHeath:
Pull out the fix to its own branch to see if it's the cause of things failing on sauce

### DIFF
--- a/fixtures/services.js
+++ b/fixtures/services.js
@@ -20,6 +20,7 @@ module.exports = {
     youthSupportNet: {
         id: 222,
         name: "Emergency Accom",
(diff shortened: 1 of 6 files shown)


danni:
Looks wrong. Are there missing files?

danni:
We also want the footer on every page.

DanielHeath:
Fixed.

DanielHeath:
Ah - I see what you mean now.

DanielHeath:
I've extracted a separate parent page so that the styleguide doesn't inherit the footer.

DanielHeath:
@danni anything else on this?

danni:
++

PR #88, opened 2015-08-25
Info box in results list

danni:
This adds a widget that can optionally display an infobox in the results listing if it's defined for the category.


DanielHeath:
++

PR #89, opened 2015-08-26
Fix some issues

danni:
- Fixes a regression with typing spaces in location names
- Some styling NQRs raised by Sarah.


DanielHeath:
++

PR #93, opened 2015-08-27
Rename classes to be consistent

danni:
The result classes had confusing names, especially now we're about to grow a map view.


.gitignore
```diff
@@ -3,5 +3,7 @@
 /src/server/webpack-stats.json
 /public/static/*.js
 /public/static/*.js.map
+/public/static/*.css
+/public/static/*.css.map
```
DanielHeath, line 7: oops - I just added this also

DanielHeath:
++

PR #94, opened 2015-08-27
Prevent attempts to merge generated code

DanielHeath:


danni:
++

danni:
Actually I take that back. I think this is wrong because it will overwrite and surprise you. I think the correct thing is to run the script, add that, which will effectively be the merge.

danni:
That way your code won't break by surprise.

DanielHeath:
This just tells git never to attempt a merge (always mark it as a conflict)

DanielHeath:
Separately we could specify a merge strategy (the script)

danni:
Sometimes it will merge fine though.

DanielHeath:
Are you happy for me to add a `.gitstrategies` or similar?

DanielHeath:
This requires git 1.7+ (which is quite old by now).

danni:
++

PR #97, opened 2015-08-27
Map view

danni:
Map view for results. Doesn't include any tests because I haven't worked out how to write tests for this yet.


kane-c:
lgtm

jscn:
lgtm

Actually this looks not that different to backbone code in places (especially the map view) which is kind of interesting.

PR #101, opened 2015-08-31
Update browsers

danni:


DanielHeath:
Yuck, but ++

danni:
Which bit is Yuck?

DanielHeath:
Sauce depending on `:8000`

DanielHeath:
Appears to have bitrotted, shall we close it?

PR #102, opened 2015-08-31
Result tile fixes

danni:
Fixes:
- Showing more than 4 related services in the results list
- Showing "invalid date"
- Showing the next open day when not tomorrow

### DIFF
--- a/src/components/AppBar.scss
+++ b/src/components/AppBar.scss
@@ -1,6 +1,7 @@
 .AppBar {
     & > div {
         display: inline-block;
+        z-index: 2;
     }
 
     h1 {
--- a/src/components/ResultListItem.js
+++ b/src/components/ResultListItem.js
@@ -45,7 +45,7 @@ class ResultListItem extends React.Component {
     /* flow:disable */
     get relatedServices(): Array<Object> {
         return this.state.relatedServices
-            .slice(0, this.props.nMoreRelatedServices);
+            .slice(0, this.props.nRelatedServices);
     }
 
     /**
@@ -56,7 +56,7 @@ class ResultListItem extends React.Component {
     get nMoreRelatedServices(): number {
         return Math.max(0,
                         this.state.relatedServices.length -
-                            this.props.nMoreRelatedServices);
+                            this.props.nRelatedServices);
     }
 
     componentDidMount(): void {
@@ -109,7 +109,7 @@ class ResultListItem extends React.Component {
                     <div>
                         {this.nMoreRelatedServices} more{' '}
                         {this.nMoreRelatedServices == 1 ?
-                            'service' : 'services'}...
+                            'service' : 'services'}…
                     </div>
                 :
                     ''
--- a/test/features/10-functional/10-basic/20-category-page.feature
+++ b/test/features/10-functional/10-basic/20-category-page.feature
@@ -50,3 +50,16 @@ Feature: Category page
         Given my location is "Melbourne VIC"
         When I visit /category/housing
         Then I should not see "Invalid date"
+
+    # FIXME: how do we mock time?
+    # Scenario: Can show opening time tomorrow
+        # Given my location is "Melbourne VIC"
+        # And today is a Tuesday
+        # When I visit /category/food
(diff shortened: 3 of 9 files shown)


src/components/AppBar.scss
```diff
@@ -1,6 +1,7 @@
 .AppBar {
     & > div {
         display: inline-block;
+        z-index: 2;
```
jscn, line 4: :(

better than 9999 I guess :p

src/components/ResultListItem.js
```diff
@@ -109,7 +109,7 @@ class ResultListItem extends React.Component {
                     <div>
                         {this.nMoreRelatedServices} more{' '}
                         {this.nMoreRelatedServices == 1 ?
-                            'service' : 'services'}...
+                            'service' : 'services'}…
```
jscn, line 112: I don't think I've seen this construct before, is it a React thing? Is it meant to be three periods or an ellipsis?

danni, line 112: It's an ellipsis. There's no HTML entities inside JS code.

src/components/AppBar.scss
danni, line 4: Does it need a comment?

test/steps/browser.js
```diff
@@ -65,6 +67,23 @@ async function clickBack(): Promise<void> {
...
 
+function pauseToDebug() {
+    return new Promise((resolve, reject) => {
+        console.log("Paused. Press any key to continue...");
+        var stdin = process.stdin;
+
+        stdin.setRawMode(true);
+        stdin.on('data', (key) => {
+            stdin.setRawMode(false);
+            if (key === '\u0003') {  // Ctrl-C
```
jscn, line 78: Nice.

test/steps/browser.js
```diff
@@ -75,6 +94,16 @@ async function thenISee(expected: string): Promise<void> {
     await assert.textIsVisible(this.driver, expected);
 }
 
+async function thenIDontSee(expected: string): Promise<void> {
+    try {
+        await assert.textIsVisible(this.driver, expected);
```
jscn, line 99: Is there some kind of timeout here? Will this give a false negative if the content is loading asynchronously? (I don't know much about promises so these not be relevant questions.)

danni, line 99: There's a Selenium implicit wait of 10s.

test/features/10-functional/10-basic/20-category-page.feature
```diff
@@ -35,3 +35,31 @@ Feature: Category page
...
+
+    Scenario: A service with 5 related services only shows 4
+        Given my location is "Melbourne VIC"
+        When I visit /category/food
+
+        Then I should see "Material Aid"
+        And I should see "Community Outreach"
+        And I should see "Crisis Accommodation"
+        And I should see "Centrelink Services"
+        And I should not see "Drug & Alcohol Counselling"
```
jscn, line 47: Am I right in thinking that this is just a hard limit of 4 and has nothing to do with the category?

test/steps/browser.js
danni, line 99: Although this might not work for the anti-case, because it thinks I'm looking for something, rather than not looking for it.

src/components/AppBar.scss
jscn, line 4: Nah, I think it's fine, just a pity that we need to hard code stuff like that because of the tools/frameworks/libraries/whatever.

test/steps/browser.js
jscn, line 99: So the test is always going to take 10s to decide the string isn't there? I guess that's acceptable but we might want to avoid making too many negative assertions like that to keep the test times down.

jscn:
++

test/features/10-functional/10-basic/20-category-page.feature
danni, line 47: Correct.

test/steps/browser.js
danni, line 99: Mmm. True.

PR #103, opened 2015-09-01
Search on service types; clean up types for ISS/GMaps

danni:
Robyn wants to be able to search on service types. Make that possible. This uncovered a few typing glitches, which I cleaned up.


kane-c:
lgtm

PR #106, opened 2015-09-01
Opening time tests; fix for this day next week

danni:
Tests for the opening times feature. Which were missing.

Also fixes showing "next Wednesday" if the next open day is a week from now.


jscn:
++

PR #107, opened 2015-09-02
Fix a 400 error caused requesting related services

danni:


src/iss.js
```diff
@@ -22,6 +22,31 @@ export type searchRequest = {
     key?: string,
 };
 
+type searchResultsLocation = {
+    name: string,
+    suburb: string,
+    state: string,
+    lat: number,
+    lon: number,
+}
```
kane-c: `;`

jscn:
Nice, ++

PR #113, opened 2015-09-05
Load more results button

danni:
A button to load more results.

### DIFF
--- a/src/constants/categories.js
+++ b/src/constants/categories.js
@@ -76,7 +76,7 @@ var categories:Array<Category> = [
         byline: "Things you might need",
         icon: icons.Things,
         search: {
-            q: 'recreation',
+            q: 'material aid',
         },
         personalisation: [
             personalisation.Location,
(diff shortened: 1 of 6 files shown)


DanielHeath:
Looks reasonable. I prefer the async / await style for promise clients but not sure if we've decided to do that everywhere or not.

danni:
@DanielHeath a specific promise? If you're referring to the UI method, that's because it's a UI method and someone needs to resolve the promise (React won't).

DanielHeath:
I was referring to `loadMore` - nothing needs to be bound to the resulting promise (unless I've misunderstood async / await)

danni:
I see, so I can just fire it and it will resolve in the future. Who will handle its exception if it throws one?

DanielHeath:
You'll get a console error but nothing else (much the same as a normal view method)

DanielHeath:
++

PR #123, opened 2015-09-08
Simplify opening times logic

DanielHeath:


danni:
++

Are you also going to handle the `now_open == null` case?

DanielHeath:
After this

PR #124, opened 2015-09-09
New Relic support

danni:
Also includes deps cleanups.


newrelic.js
```diff
@@ -0,0 +1,23 @@
...
+ * New Relic agent configuration.
+ *
+ * See lib/config.defaults.js in the agent distribution for a more complete
+ * description of configuration variables and their potential values.
+ */
+exports.config = {
+    /**
+    * Array of application names.
+    */
+    app_name: [`Ask Izzy (${process.env.ENVIRONMENT})`],
```
DanielHeath, line 11: default to 'dev' or something? (maybe check hostname?)

DanielHeath:
++

newrelic.js
danni, line 11: `ENVIRONMENT` will be set anywhere the NEW_RELIC key is set.

PR #127, opened 2015-09-09
Service object

danni:
Turns service into an class we can add methods to.

Returns service provisions on the results list page and adds sibling services to the service details page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,27 @@ help:
 
 ## Running the dev server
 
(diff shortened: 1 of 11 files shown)


DanielHeath:
Nice :)

DanielHeath:
++ once tests pass

danni:
@DanielHeath ++?

DanielHeath:
++

PR #130, opened 2015-09-10
Turns out there’s important stuff in icons path-styles

DanielHeath:
Turns out we were discarding the style property of our SVG icons, which dropped (eg) stroke.


danni:
Are we sure about this? It seems like it will introduce random colours. Perhaps we should determine the icons that have a fill style and then pull that out into SCSS?

DanielHeath:
Color can matter (eg the logo has a white 'izzy'). I've overridden it with !important where necessary.

danni:
I agree colour matters. I was just wondering since we generate class names we should write this out via CSS styles using parameters.

DanielHeath:
Right, but then we'd need to give the different `path` elements different classes within the SVG. 
Could be done but sounds likely to result in bees to me.

danni:
Okay ++

danni:
This has broken the chevron colours.

PR #133, opened 2015-09-10
Inline the main logo so it doesn't load after the rest of the page

DanielHeath:


src/pages/HomePage.js
```diff
@@ -31,11 +31,8 @@ class HomePage extends React.Component {
             <div className="HomePage">
                 <div className="header">
                     <div className="branding-container">
-                        <img
-                            className="branding-icon"
-                            src="/static/askizzy-icon-logotype.svg"
-                            alt="Ask Izzy"
-                        ></img>
+                        { <icons.Logotype className="branding-icon" /> }
```
danni: Do you need the `{}` ?

danni: Also can we replace `.branding-icon` by `.Logotype`?

PR #134, opened 2015-09-10
Add checks that environment variables are set before attempting tests…

DanielHeath:
…/dev-server

### DIFF
--- a/script/dev-server
+++ b/script/dev-server
@@ -3,6 +3,11 @@ set -euo pipefail
 IFS=$'\n\t'
 PATH=`npm bin`:$PATH
 
+if [ -z "${ISS_URL:=}" ] ; then
+    echo "ISS_URL not set; cannot run."
+    exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


script/dev-server
```diff
@@ -3,6 +3,11 @@ set -euo pipefail
 IFS=$'\n\t'
 PATH=`npm bin`:$PATH
 
+if [ -z "${ISS_URL:=}" ] ; then
+    echo "ISS_URL not set; cannot run."
+    exit 1
+fi
```
danni, line 9: Isn't this already in `server.js`?

DanielHeath, line 9: Nodemon doesn't quit if the process crashes instantly up-front, so then you have to ctrl-c `nodemon`.

Also, the output gets surrounded by a bunch of colorized nodemon stuff which is hard to read, and the focus is stolen by the browser window opening.

danni:
Okay ++

PR #135, opened 2015-09-10
When printing, show SVGs as solid black.

DanielHeath:
e.g. the primary logo includes white text for ‘izzy’
which doesn’t show up when printing.


danni:
++

PR #141, opened 2015-09-12
Service provisions redesign

danni:
This is a complete redesign of the service provisions work to make it much more descriptive.
- Allow finding keywords within sentences without writing complicated regexes.
- Introduce common regexes for optional hyphens and space normalisation.
- Yaml file of test cases that are fed into Mocha as unit tests.
- anyOf, allOf, not, and keywords allow for descriptive matching of provisions.
- Much more OO, so easy to write even more complicated rules if we need to.

### DIFF
--- a/script/test
+++ b/script/test
@@ -15,4 +15,10 @@ fi
 # build assets if this isn't Travis (which has already)
 ${CI:=false} || ./script/build-assets
 
-mocha --timeout 0 --compilers js:babel/register test/*.js "$@"
+if [ -z "$@" ]; then
+    TESTS=test/*.js
+else
+    TESTS="$@"
+fi
+
+mocha --timeout 0 --compilers js:babel/register "$TESTS"
--- a/script/test
+++ b/script/test
@@ -15,10 +15,4 @@ fi
 # build assets if this isn't Travis (which has already)
 ${CI:=false} || ./script/build-assets
 
-if [ -z "$@" ]; then
-    TESTS=test/*.js
-else
-    TESTS="$@"
-fi
-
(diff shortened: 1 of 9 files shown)


script/test
```diff
@@ -15,4 +15,10 @@ fi
 # build assets if this isn't Travis (which has already)
 ${CI:=false} || ./script/build-assets
 
-mocha --timeout 0 --compilers js:babel/register test/*.js "$@"
+if [ -z "$@" ]; then
+    TESTS=test/*.js
```
DanielHeath: Doesn't mocha default to `tests/*.js` anyways?

src/ServiceProvisions.js
```diff
@@ -0,0 +1,173 @@
...
+"use strict";
+
+import _ from 'underscore';
+
+function escapeRegex(s) {
+    return s.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
+};
+
+/**
+ * ServiceProvision aka provides:
```
DanielHeath: aka `provider`?

src/ServiceProvisions.js
```diff
@@ -0,0 +1,173 @@
+/* @flow */
```
DanielHeath, line 1: I'm concerned that `src/*` is getting pretty long; is there a sensible folder for this?

DanielHeath, line 1: Is this for generating searches? Displaying stuff?

danni, line 1: Dunno. Probably not. It's for generating the list of service provisions.

DanielHeath:
++

PR #142, opened 2015-09-14
Reinstate "Inline the main logo so it doesn't load after the rest of …

DanielHeath:
…the page"

Currently the header appears after the rest of the page has loaded.
This change renders the SVG inline which ensures the logo appears with the text.

This reverts commit 16db07b3024707242fda04cb790e9a4c16205a9c.


danni:
Is it going to crash Chrome still?

danni:
Also needs the SVG CSS directive to align to the pixel grid. As the sharp edges are anti-aliased on my screen.

danni:
Okay, Chrome works for me now. Maybe it was to do with the sizing/trying to resize it in devtools dynamically. Or maybe I'd run out of memory before I rebooted yesterday.

danni:
One other difference is the Izzy is now higher than the Ask.

danni:
screenshot from 2015-09-14 10-49-00
screenshot from 2015-09-14 10-48-55

danni:
Your other option is to base64 inline it. Which I believe is a thing Webpack can do for you.

DanielHeath:
@isaacwilliams I'll ship this once I get the trimmed version of the logo.

DanielHeath:
++

PR #143, opened 2015-09-14
Demographics & gender

danni:
N.B. This breaks the styling of the radio icons. Fixing this now in material-ui.

### DIFF
--- a/src/components/TransportTime.js
+++ b/src/components/TransportTime.js
@@ -28,15 +28,29 @@ class TransportTime extends React.Component {
             object,
         } = this.props;
 
-        return (
-            <div className="TransportTime">
-                <icons.Walk className="ColoredIcon brand-text-dark" />
-                <span className="time">? mins</span>&nbsp;
-                <span className="location">
-                    {titleize(this.props.object.location.suburb)}
-                </span>
-            </div>
-        );
+        if (object.location.point) {
+            return (
+                <div className="TransportTime">
+                    <icons.Walk className="ColoredIcon brand-text-dark" />
+                    <span className="time">? mins</span>&nbsp;
+                    <span className="location">
+                        {titleize(this.props.object.location.suburb)}
+                    </span>
+                </div>
+            );
+        } else {
+            /* This is a confidential location, we can't show any
+             * transport time*/
+            return (
+                <div className="TransportTime">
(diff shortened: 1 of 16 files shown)


danni:
P.R. to fix customising of RadioButtons https://github.com/callemall/material-ui/pull/1639

src/components/TransportTime.js
```diff
@@ -28,15 +28,29 @@ class TransportTime extends React.Component {
...
+                    <span className="location">
+                        {titleize(this.props.object.location.suburb)}
+                    </span>
+                </div>
+            );
+        } else {
+            /* This is a confidential location, we can't show any
+             * transport time*/
+            return (
+                <div className="TransportTime">
```
DanielHeath, line 45: I'd generally prefer to split this kind of conditional into separate components.

src/pages/personalisation/BaseQuestion.js
```diff
@@ -52,13 +60,25 @@ class BaseQuestion extends React.Component {
...
 
-    onAnswerTouchTap(answer: string, ...rest: any): void {
-        storage.setItem(this.props.name, answer);
+    /**
+     * triggerNext:
+     *
+     * Trigger next page after a 500ms debounce.
+     */
+    /*::__(){`*/@debounce(500)/*::`}*/
+    triggerNext(): void {
```
DanielHeath, line 69: Could we use css transitions to give this visual feedback?

src/components/TransportTime.js
danni, line 45: Well it's more the component can't do anything. Unless you wanted the component to return ''?

DanielHeath, line 45: I'd probably have it return `<ConfidentialLocation/>` or `<TransportTimeToPoint/>`

danni, line 45: That is too much abstraction IMO. It's simply a case of "can't do what I'm asked, falling back"

src/pages/personalisation/BaseQuestion.js
danni, line 69: We are. However we also need to have the time to see that.

src/components/TransportTime.js
DanielHeath, line 45: Fair enough; maybe if the component were getting huge.

DanielHeath:
++

PR #144, opened 2015-09-15
Maps links open in the same tab

DanielHeath:


danni:
++

danni:
Of course, if the user now hits back, the app will break, because it will have lost context.

danni:
The other option is to just add a `title` or `aria-label` attribute "Opens in new tab"

DanielHeath:
Yeah - not 100% sold on writing our own browser-history wrapper.

DanielHeath:
What context will be missing? SessionStore will still be there, right?

DanielHeath:
I've switched back to opening a new tab and adding an `aria-label`

src/components/Address.js
```diff
@@ -48,6 +48,7 @@ class Address extends React.Component {
                 </ScreenReader>
                 <a
                     target="_blank"
+                    aria-label="Open google maps in a new tab"
```
danni: Google Maps is a proper noun.

danni:
Otherwise ++

PR #147, opened 2015-09-16
Wrap header logo in a div so that safari shows the top padding correctly

DanielHeath:


danni:
++

danni:
The header on other pages is also quite high on Safari.

PR #150, opened 2015-09-16
Crisis accommodation fix; leave question out when you need crisis accommodation

danni:


src/pages/personalisation/Location.js
```diff
@@ -79,6 +79,10 @@ class Location extends React.Component {
         return storage.getItem('location');
     }
 
+    static showQuestion(): boolean {
+        return true;
```
DanielHeath, line 83: Isn't this inherited?

danni, line 83: No.

DanielHeath:
++, would like to see BaseQuestion split out so Location can inherit later.

PR #154, opened 2015-09-17
Age personalization now has the correct age ranges and hooks into ISS

DanielHeath:


src/iss.js
```diff
@@ -293,6 +294,35 @@ export async function request(path: string, data: ?searchRequest): Object {
...
+     * Encode data into the URI ourselves
+     * until we can work around
+     * https://github.com/jedmao/iso-http/issues/2
+     * No flow until comprehensions are supported
+    */
+    var serialized = ""
+    if (data) {
+        // Flow can't tell that `data` isn't null inside a closure
+        var _data = data;
+        serialized = Object.keys(data).map(function (key) {
```
danni: Use an ES6 closure?

src/iss.js
```diff
@@ -293,6 +294,35 @@ export async function request(path: string, data: ?searchRequest): Object {
     urlobj.auth = urlobj.search = urlobj.querystring = urlobj.query = null;
     url_ = url.format(urlobj);
 
+    /*
+     * Encode data into the URI ourselves
+     * until we can work around
+     * https://github.com/jedmao/iso-http/issues/2
+     * No flow until comprehensions are supported
+    */
+    var serialized = ""
```
danni: Missing ;

src/iss.js
DanielHeath: I think the problem is that flow can't determine that the closure never leaks outside the `if` block.

DanielHeath: In any case, an arrow function doesn't help with the flow error & we still need to use `return` because of the `/* ::` */` stuff.

src/iss.js
```diff
@@ -293,6 +294,35 @@ export async function request(path: string, data: ?searchRequest): Object {
...
+        var _data = data;
+        serialized = Object.keys(data).map(function (key) {
+            if (!_data[key]) {
+                throw new Error(`Invalid value specified for ${key}`);
+            }
+            /* ::` */
+            if (_data[key][Symbol.iterator]) {
+                return [
+                    for (v of _data[key])
+                        key + '=' + encodeURIComponent(v)
```
danni: Weird indenting. Should all be inline. Also use a template string :)

src/iss.js
```diff
@@ -293,6 +294,35 @@ export async function request(path: string, data: ?searchRequest): Object {
...
+                ].join('&');
+            }
+            /* ::` */
+            return `${key}=${encodeURIComponent(_data[key])}`;
```
danni: `else` (makes it easier to read the flow control).

danni:
++ with style fixes. Assuming the tests pass and you're positive it won't break my demo :-P

src/iss.js
DanielHeath: `else` is hard for flow because we need "/\* ::`*/" around the`if`

DanielHeath: oh wait, it's fine since we don't need `else {`

danni: Move the `/*::` inside the array so flow thinks it returns []. Have a look at other uses of generator comprehension in Izzy.

src/pages/personalisation/Questions.js
```diff
@@ -27,19 +27,43 @@ export class SleepTonight extends BaseQuestion {
...
     // flow:disable
     static defaultProps = {
         name: "age",
         question: "How old are you?",
-        answers: [
-            "26 or younger",
-            "27 to 54",
-            "55 or older",
-        ],
+        answers: Object.keys(ageGroups),
```
danni, line 62: Nice :)

src/iss.js
DanielHeath: Can't do that with a template string inside the comments.

danni: Okay.

DanielHeath: Turns out there was a really nice way to do this by not repeating myself.

PR #155, opened 2015-09-17
Don't try to jscs flowtype interfaces

DanielHeath:


danni:
Why?

DanielHeath:
I can't think of a good reason any more. 

I was trying to resolve some flow errors with `if (_.isArray(thing))` not casting `thing` to array inside the if block.

I added an underscore lib definition cribbed from github.com/solnetdigital/flow-interfaces-underscore (which didn't end up helping) but it failed JSCS.

danni:
I think you can just tell jscs to skip one file.

DanielHeath:
Sure, but I also can't think of a good reason to jscs files which don't have any javascript in them.

danni:
Strictly speaking it is JS.

DanielHeath:
I'm not aware of a JS engine that'll use it; babel strips it out before doing anything else.

danni:
No JS engine uses any code style :-P

DanielHeath:
Fair

PR #159, opened 2015-09-17
Empty arrays are truthy in JS

DanielHeath:


src/components/Eligibility.js
```diff
@@ -24,7 +24,7 @@ class Eligibility extends React.Component {
         var eligibleMarkup;
         var ineligibleMarkup;
 
-        if (eligibilityInfo || catchment) {
+        if (eligibilityInfo.length || catchment.length) {
```
danni, line 34: `_.isEmpty` is more lucid IMO

danni:
Tests?

danni:
++

However I think we should break the service details page out into a number of features.

DanielHeath:
It's definitely getting overly long.

PR #165, opened 2015-09-21
Flowtype doesn’t typecheck React.Element, use ReactElement

DanielHeath:


src/components/CollapsedPhones.js
```diff
@@ -52,10 +52,9 @@ class CollapsedPhones extends React.Component {
                     {extras}
                 </div>
             );
-
         } else {
             // No phones
-            return null;
+            return <div className="CollapsedPhones" />;
```
danni: `return ''`

src/components/CrisisLineItem.js
```diff
@@ -52,7 +52,7 @@ class CrisisLineItem extends React.Component {
             }
         }
 
-        return null;
+        return  <div className="CrisisLineItem"/>;
```
danni: `return ''`

danni:
Otherwise ++

PR #166, opened 2015-09-21
Use factories instead of fixtures

DanielHeath:


test/support/dictionary.js
```diff
@@ -35,26 +36,80 @@ function tableConverter(str: string, done: callback): void {
...
+    done(null, Service(serviceProps));
 }
 
 var dictionary = new Yadda.Dictionary()
     .define('LATITUDE', /(\d+.\d+[NS])/, latitudeConverter)
     .define('LONGITUDE', /(\d+.\d+[EW])/, longitudeConverter)
     .define('lines', /([^\u0000]*)/, linesConverter)
-    .define('table', /([^\u0000]*)/, tableConverter);
+    .define('table', /([^\u0000]*)/, tableConverter)
+    .define('service', /([^\u0000]*)/, serviceConverter);
```
danni: Style: Put `;` on own line. Makes diffs clearer.

test/features/10-functional/10-basic/40-service-details.feature
```diff
@@ -17,8 +17,25 @@ Feature: Service details page
...
+            fax      | (03) 5555 5555 |
+            mobile   | 0477 777 777   |
+            phone    | 0477 777 777   | (really a mobile)
+            freecall | 1300 111 111   |
+            tty      | (03) 9999 9999 |
+        * Emails
+            email                 | comment
+            ==============================================
+            reception@service.org | Reception
+        * Web: https://service.org
```
danni, line 36: Should we actually just use Yaml?

danni:
How to I use this to create a set of services?

DanielHeath:
@danni: You can't (yet), though most of our scenarios refer to a single service.

Options for this:
- Extend `Given a service with` to `Given a service named "" with` and use the name to refer back to it.
- Write a step which mocks the thing you actually need (`Given the following 'housing' services, When I search for 'housing'`)

danni:
Do we want to do that first?

danni:
Happy to merge after #168 -- as there's style work here I've not complained about because the linter will do it now.

DanielHeath:
#168 should definitely come first. 

I'd like to get this in and start using it for the tests where it's easy to do so, then see if we still have trouble with fixtures.

danni:
++ after conflict merge.

test/features/10-functional/10-basic/40-service-details.feature
DanielHeath, line 36: Can you do key/value pairs on a single line with YAML? 
This phone example would be _long_ otherwise.

PR #167, opened 2015-09-21
Add personalisation link when bad location was set

pfctdayelise:
Detect when ISS gives a 402 error specifically which indicates a bad location. Closes https://trello.com/c/NYgRDR1i/186-unable-to-change-the-location

Added test for this case, and also a test for the case of what we show with other errors, just for completeness.


src/pages/ResultsListPage.js
```diff
@@ -47,26 +45,22 @@ class ResultsListPage extends BaseResultsPage {
                             <div>Searching...</div>
                     }
                     secondaryText={
-                        this.state.meta ?
+                        this.state.statusCode == 402 ?
```
danni, line 45: `&& this.state.error`. It might have failed to parse.

src/pages/ResultsListPage.js
```diff
@@ -84,6 +78,27 @@ class ResultsListPage extends BaseResultsPage {
         );
     }
 
+    errorMessage(): React.Element {
+        var message = this.state.error;
+        return <p className="errorMessage">{ message }</p>;
```
danni: Style: no space between `{ }`

src/pages/ResultsListPage.js
```diff
@@ -84,6 +78,27 @@ class ResultsListPage extends BaseResultsPage {
         );
     }
 
+    errorMessage(): React.Element {
+        var message = this.state.error;
```
danni, line 77: Style: Newline here

test/support/mock_iss/server.js
```diff
@@ -248,6 +248,18 @@ app.get('/api/v3/search/', (req, res) => {
                 services.domesticviolence,
             ],
         });
+    } else if (req.query.q.match(/elasticsearch/)) {
+        res
```
danni, line 258: Make this something more explicit. i.e. elasticsearch unavailable.

test/support/mock_iss/server.js
```diff
@@ -248,6 +248,18 @@ app.get('/api/v3/search/', (req, res) => {
...
+            .status(503)
+            .json({
+                error_message: 'The Elasticsearch service is unavailable.',
+            });
+    } else if (req.query.area == "carlt") {
```
danni: Should be higher in the branches. Else a valid search will override it, and it's effectively always an error.

src/pages/ResultsListPage.js
pfctdayelise, line 45: Because the error and statusCode are set simultaneously I think this is ok as-is.

danni:
++

PR #169, opened 2015-09-22
Fix styling bug on crisis lines in prod

pfctdayelise:
This occurred when CSS was minified (as in prod). Can be reproduced by running prod-server. 

elem.displayName is safer than elem.type.name for this purpose!

Fixes https://trello.com/c/GhHSiFCS/194-grey-line-appears-between-crisis-lines-on-ios-and-windows-mobile


danni:
++

PR #170, opened 2015-09-22
Address component should not duplicate the word 'level' from ISS

DanielHeath:
This isn't quite there yet - need to get a product decision about what order to show unit, level & building (as well as confirm whether the `flat_unit` field contains just a number, or a number + description.


test/features/10-functional/10-basic/40-service-details.feature
```diff
@@ -54,8 +54,21 @@ Feature: Service details page
         Then I should not see "You are ineligible if"
 
     Scenario: The address is a link to google maps
-        When I visit /service/866464
-        Then I can get to google maps by clicking "33 Elizabeth Street Richmond VIC 3121"
+       Given A service with:
+        ----------------------------------------------
+        * Location{}
```
danni: Can we use something other than `{}` like `:` ?

test/support/dictionary.js
```diff
@@ -26,7 +26,37 @@ function linesConverter(str: string, done: callback): void {
     done(null, str.split("\n"));
 }
 
+// Converts a key from the format used in features to an object key
+function sanitizeKey(key: string): string {
```
danni, line 30: Is this really `_.string.slugify`?

test/support/dictionary.js
```diff
@@ -26,7 +26,37 @@ function linesConverter(str: string, done: callback): void {
...
+     * Key | Value
+     */
+    var downcaseKey = arr => [sanitizeKey(arr[0]), arr[1]];
+
+    return _.object(
+        lines.map(
+            line => line.split("|").map(
+                cell => cell.trim()
+            )
+        ).map(downcaseKey)
```
danni: You can combine this into one map.

test/support/dictionary.js
```diff
@@ -69,14 +94,12 @@ function serviceConverter(str: string, done: callback): void {
     var serviceProps = {};
     var currentTableLines, // eg "Emails"
         currentTable,      // the lines which constitute a table
+        parseFn,// how should currentTableLines be parsed
```
danni: Weird indenting.

test/features/10-functional/10-basic/40-service-details.feature
DanielHeath: We need a way to do objects, arrays and values - I've used the following for now:
`Key: Value` for bare values
`Key: (nada)` for empty values
`Key:` for arrays
`Key{}` for objects

DanielHeath: Alternatively we could switch on the presence of header.

test/support/dictionary.js
DanielHeath, line 30: Doesn't `slugify` use dashes instead of underscores?

DanielHeath:
@danni ++?

danni:
++ :+1:

PR #172, opened 2015-09-23
Fix chevron sizing across all components

DanielHeath:


danni:
Given this has broken many times should we have a test for it?

danni:
++ for what's here anyway

PR #176, opened 2015-09-24
Split large vendored libraries out

DanielHeath:
This puts react (600kb) and material-ui (500kb) into their own bundle (along with a few of the other larger libraries).


danni:
++

PR #177, opened 2015-09-24
Add transport time/directions link to service detail view

pfctdayelise:
Closes https://trello.com/c/rxzpKN7H/209-no-directions-link-time-estimate-in-detail-view

Added 'isConfidential' flag to services, based on if they have a location.point value or not.

Confidential services have no transport/directions info included on their individual page. Also now the first line of their address will read "Confidential location".

Non-confidential services have the travel icon and time estimate same as is currently in the results listing view, but without the suburb (as the address is right above), and a "Get directions" link which is for Google Maps directions from "Current Location" to location.point.

### DIFF
--- a/src/components/Address.js
+++ b/src/components/Address.js
@@ -6,6 +6,17 @@ import { titleize } from "underscore.string";
 import ScreenReader from "./ScreenReader";
 
 class Address extends React.Component {
+    // flow:disable not supported yet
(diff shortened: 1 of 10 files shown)


pfctdayelise:
Conflicting with @DanielHeath's work in another branch, closing this for now.

PR #180, opened 2015-09-28
When setting directions, use walking mode & valid startpoint

DanielHeath:
Google maps doesn’t actually accept the string “Current Location”.

### DIFF
--- a/src/components/GoogleMapsLink.js
+++ b/src/components/GoogleMapsLink.js
@@ -12,19 +12,19 @@ class GoogleMapsLink extends React.Component {
                 <div>Link text</div>
             ),
             to: new Location(fixtures.ixa.location),
+            from: "",
         },
     };
 
     googleMapsUrl(): string {
         var toAddr = this.props.to;
 
-        // FIXME: this.props.from
-        var start = "Current+Location";
+        var start = encodeURIComponent(this.props.from || "");
         var query = encodeURIComponent(
             `${toAddr.streetAddressLine1()} ${toAddr.streetAddressLine2()}`
         );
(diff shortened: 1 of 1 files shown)


src/components/GoogleMapsLink.js
```diff
@@ -12,19 +12,19 @@ class GoogleMapsLink extends React.Component {
...
+            from: "",
         },
     };
 
     googleMapsUrl(): string {
         var toAddr = this.props.to;
 
-        // FIXME: this.props.from
-        var start = "Current+Location";
+        var start = encodeURIComponent(this.props.from || "");
```
danni: Use `defaultProps` and add a comment explaining the meaning of ""

src/components/GoogleMapsLink.js
```diff
@@ -12,19 +12,19 @@ class GoogleMapsLink extends React.Component {
...
         var query = encodeURIComponent(
             `${toAddr.streetAddressLine1()} ${toAddr.streetAddressLine2()}`
         );
 
-        return `https://maps.google.com/?saddr=${start}&addr=${query}`;
+        return `https://maps.google.com/?directionsmode=walking&saddr=${start}&daddr=${query}`;
```
danni, line 32: Can we do walking or tramming?

DanielHeath, line 32: You can only specify one of them; walking seemed like the safer option.

danni, line 32: Let's add a FIXME to do it based on whichever we're showing in the travel time widget once that works.

PR #181, opened 2015-09-28
Remove personalization other than location from all non-housing categories

DanielHeath:
At Dianas request.


danni:
We still want location.

danni:
Ignore me. ++

PR #184, opened 2015-09-29
Add timeout for promises so the tests don’t hang

DanielHeath:


danni:
++

DanielHeath:
So, this is timing out because `Then I should see the results` sends _so_ many queries to saucelabs that a single step can take over 80 seconds. I don't think the solution is to increase the timeout...

PR #187, opened 2015-09-30
Fix flow bugs

DanielHeath:


script/typecheck
```diff
@@ -11,6 +11,24 @@ function check {
     echo "Good"
 }
 
+badFlow=$( grep '@flow' -n -r src test | grep -v :1: | sed s/:.*// || echo "" )
```
danni: Use `read -n 1`

DanielHeath:
Opened https://github.com/facebook/flow/issues/902

script/typecheck
```diff
@@ -11,6 +11,26 @@ function check {
     echo "Good"
 }
 
+echo "${BOLD}Checking for missing @flow comments${END}"
+echo "Good"
+
+missingflow=""
```
danni: you can literally use the binaries true/false.

So `HAVE_FLOW=true`

script/typecheck
```diff
@@ -11,6 +11,26 @@ function check {
...
+for file in $( git ls-files '**/*.js' ); do
+    read flowtype < "$file"
+    if test "x$flowtype" = 'x/* flow:disable */'; then
+        echo "Skipping disabled file $file" > /dev/null
+    elif test "x$flowtype" != 'x/* @flow */'; then
```
danni, line 22: `HAVE_FLOW=HAVE_FLOW && test ...`

script/typecheck
```diff
@@ -11,6 +11,26 @@ function check {
...
+        missingflow="true"
+        cat << EOF
+$file is not annotated for flowtype (must be the first line)! - got "$flowtype"
+EOF
+    fi
+done
+
+if [ "$missingflow" = "true" ] ; then
```
danni: `if ! HAVE_FLOW; then`

src/iss.js
```diff
@@ -128,9 +129,13 @@ export function mungeUrlQuery(url_: string, data: Object): string {
     return url_ + joiner + serialized;
 }
 
-export async function request(path: string, data: ?searchRequest): Object {
+export async function request(
+    path: string,
+    data: ?searchRequest
+): Promise<Object> {
```
danni, line 135: I think now `async` functions no longer need to return `Promise<blah>` and can just return `blah`.

src/iss.js
```diff
@@ -128,9 +129,13 @@ export function mungeUrlQuery(url_: string, data: Object): string {
...
     var url_: string = ISS_URL || process.env.ISS_URL;
 
+    /* flow:disable url module does something truly bizarre to flow */
```
danni: ... Is it that we're redefining `url` somewhere? That's typically the problem.

src/iss.js
```diff
@@ -295,6 +300,7 @@ export class Service {
         }
 
         var request_: searchRequest = {
+            q: "",
```
danni: Not correct.

danni: You want to define `q` as optional. Sending `q` in the query will change what happens.

src/storage.js
```diff
@@ -15,16 +15,23 @@ class Storage {
         sessionstorage.setItem(key, obj);
     }
 
-    static getJSON(key: string): ?(Object|Array) {
+    static getJSON(key: string): any {
```
danni, line 18: Why?

src/storage.js
```diff
@@ -15,16 +15,23 @@ class Storage {
...
-            return JSON.parse(this.getItem(key));
+            return JSON.parse(item);
         } catch (error) {
             console.error(error);
             return null;
         }
     }
 
-    static setJSON(key: string, obj: Object|Array): void {
+    static setJSON(key: string, obj: mixed): void {
```
danni, line 34: Why?

webpack/utils/write-stats.js
```diff
@@ -26,8 +28,7 @@ module.exports = function writeStats(stats) {
         var unorderedChunks = _(allChunkNames).difference(chunkOrdering);
         var sortedChunkNames = chunkOrdering.concat(unorderedChunks);
 
-        for (var name of sortedChunkNames) {
-
+        for (var name of sortedChunkNames) if (json.assetsByChunkName[name]) {
```
danni: One block per line please. This sort of construct just leads to mistakes.

danni:
So if Flow now supports `let` and `const`, should we switch to using let/const. Since I think I've specified in eslint to apply `let` semantics to all uses of `var`, it should be easy to `sed` them all to `let` and swap eslint over to denying `var` and requiring `let`/`const`. We can then go back to adding `const` where appropriate.

danni:
Using more `const` would probably also help with things like where we accidentally override the URL module with a variable called URL. Although I would expect there's already a eslint rule for shadow declarations we could turn on.

DanielHeath:
That sounds good to me, but I'm keenly aware of how fast this PR bitrots (any new file in master causes it to fail the merge build). Happy to do that separately?

script/typecheck
DanielHeath, line 22: assignment inside an `if`?

danni, line 22: Oh, I missed the second line. Then you can just do `HAVE_FLOW=false`.

src/storage.js
DanielHeath, line 18: `JSON.parse` is typed `any` in flow - it's valid to do `JSON.parse('"a string"')`

src/storage.js
DanielHeath, line 34: The union type doesn't work in function declarations but this file had never been typechecked.

src/iss.js
DanielHeath: I've filed a bug with flow - `url.format(url.parse('a url'))` doesn't work.

src/storage.js
danni, line 34: I don't think that's correct. According to the docs:

```
Sometimes, multiple signatures are not needed to express overloading: the signatures can 
be coalesced using union types. Flow provides the following syntax for union types:

T1 | .. | Tn

is the union of types Ti. Union types are available for general use.
```

src/iss.js
danni: URL in the comment?

src/storage.js
DanielHeath, line 34: As discussed: If you can get it working I'll be super happy, but I've sunk a significant bit of time into it without any progress.

DanielHeath:
Failed due to hitting the saucelabs tunnel limit; other builds passed.

DanielHeath:
Previous build passed, and latest commit is only a comment change (and passed the typecheck).

PR #189, opened 2015-10-05
Branding updates

DanielHeath:


src/constants/categories.js
```diff
@@ -15,7 +15,7 @@ class Category {
     byline: string;
     icon: ReactComponent;
     search: iss.searchRequest;
-    info: ?ReactElement;
+    info: ?string;
```
danni: :(

danni: `ReactElement|string`

src/pages/BaseResultsPage.js
```diff
@@ -50,24 +50,6 @@ class BaseResultsPage extends BaseCategoriesPage {
             objects = [];
         }
 
-        /* splice in an infobox (if it exists) after the first non-crisis
-         * service */
-        try {
```
danni, line 55: I would comment this out with a reason why.

danni:
You will also need to drop the test for the infobox.

danni:
Have the tests run on the most recent commit? I don't see a tick. ++ assuming it passes.

PR #191, opened 2015-10-05
Move icons into the vendor bundle as they change infrequently.

DanielHeath:


danni:
icon bundle?

DanielHeath:
We could (or just rename to `longlived` and `shortlived`). Adding more separate fetches might not be a win.
Does prod speak `HTTP/2`?

PR #193, opened 2015-10-06
Don't double quote the message for "zero results"

DanielHeath:


danni:
Should we actually be removing the semantic quotes? Especially if the did you mean becomes `zero results` without quotes.

danni:
So technical ++, I just wonder if it will confuse people, because the quotes have semantic meaning.

DanielHeath:
I've brought this up in https://trello.com/c/bSdq8fDP/151-defect-quotes-not-stripped-from-search-term-and-are-presented-in-the-search-results-as-parking-fines-double-lot-of-them-fix-not- - hopefully we get a good answer from someone product-ey

PR #194, opened 2015-10-07
Upgrade libs

DanielHeath:
Lots of separate commits to help track down any CI issues


danni:
++ for all of this but can we write a test for the history bug?

DanielHeath:
Will do - this doesn't solve the history card on its own anyways as we want the back button to take you to `/` even when you've come from (eg) an email deep-link.

PR #195, opened 2015-10-07
Avoid checking URL too early after a click

DanielHeath:
We may not have transitioned to the new page yet when we check currentUrl

### DIFF
--- a/test/steps/browser.js
+++ b/test/steps/browser.js
@@ -72,7 +72,38 @@ async function clickBack(): Promise<void> {
         .click();
 }
 
+/**
+ * Wait for the document to be ready (including completing any AJAX requests).
+ *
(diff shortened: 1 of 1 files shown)


danni:
I'm surprised we need this TBH. Since there's an implicit wait in Selenium. What's the use case? It probably needs a comment.

DanielHeath:
So, this 'fixes' it but not for the reasons I had thought (it's basically a `sleep`). 

The underlying problem is that checking the URL doesn't wait for the browser to finish executing JS. As we use history.pushState, a better solution might be to force the use of JS to fetch the URL (since that will add a new event to the browsers queue).

DanielHeath:
++?

DanielHeath:
nvm, this is not ready

DanielHeath:
++?

danni:
++

PR #198, opened 2015-10-08
Fix navigation

DanielHeath:


src/pages/PersonalisationWizardPage.js
```diff
@@ -21,6 +21,15 @@ class PersonalisationWizardPage extends BasePersonalisationPage {
         var subpage = this.state.subpage - 1;
 
         if (subpage < 0) {
+            /*
+             * FIXME: If we use a real URL for each subpage,
```
DanielHeath: @danni I'd love to hear your thoughts on this approach to getting the back/forwards to work nicely in personalisation.

I think it'll let us use regular history while still preserving the desired behavior of having the `back` button skip over the personalisation questions.

DanielHeath:
++?

src/client.js
```diff
@@ -10,7 +10,20 @@ import routes from "./routes";
 // For onTouchTap: see https://github.com/zilverline/react-tap-event-plugin
 injectTapEventPlugin();
 
+/*
+ * If at any point there isn't a meaningful 'back',
+ * go to the app homepage instead.
+ */
 let history = createBrowserHistory();
+const goBack = history.goBack;
```
danni: Can we do this properly by inheriting the history class from ReactRouter ?

src/pages/BaseResultsPage.js
```diff
@@ -107,9 +107,15 @@ class BaseResultsPage extends BaseCategoriesPage {
...
+                const sep = this
+                    .props
+                    .location
+                    .pathname
+                    .endsWith("/") ? "" : "/";
+
                 this.history.replaceState(
                     null,
-                    this.props.location.pathname + "/personalise",
+                    this.props.location.pathname + sep + "personalise",
```
danni: Template string?

test/steps/browser.js
```diff
@@ -26,6 +26,15 @@ module.exports = (function() {
               unpromisify(doSearchAndEnter))
         .when("I click on the search icon", unpromisify(clickSearchIcon))
         .when("I click back from the title bar", unpromisify(clickBack))
+        .when(
+            "I click back from the browser UI",
+            unpromisify(clickBrowserBack),
+        )
+        .when(
+            "I click back from the top bar",
+            unpromisify(clickAppBarBack),
```
danni: This already exists I think?

src/pages/PersonalisationSummaryPage.js
```diff
@@ -10,23 +10,15 @@ import icons from "../icons";
...
-            this.props.history.goBack();
-        } else {
-            this.setState({subpage: 0});
-        }
+        this.props.history.goBack();
     }
 
     nextStep(): void {
-        this.setState({subpage: 0});
+        this.props.history.goBack();
```
danni, line 17: Can't we just call goBack from the `NavigationMixin` ?

test/steps/browser.js
DanielHeath: Indeed it does, thanks.

src/pages/PersonalisationSummaryPage.js
DanielHeath, line 17: `NavigationMixin` was replaced with `History` mixin in the new react-router, which is what I've used here.

src/client.js
DanielHeath: It's not a class, but I found a cleaner way to do it nonetheless.

src/client.js
```diff
@@ -10,9 +10,55 @@ import routes from "./routes";
 // For onTouchTap: see https://github.com/zilverline/react-tap-event-plugin
 injectTapEventPlugin();
 
-let history = createBrowserHistory();
+/*
+ * If at any point there isn't a meaningful 'back',
+ * go to the app homepage instead.
+ */
+
+function History() {
```
danni, line 22: Why not use a real class?

PR #199, opened 2015-10-13
Fix a bug that prevented ‘also at this location’ displaying

DanielHeath:


DanielHeath:
The errors this generates are silently swallowed :(

DanielHeath:
Tests to come.

danni:
++

PR #209, opened 2015-10-21
Remove extraneous !important, fix some scaling issues

danni:


src/components/StaticMap.js
```diff
@@ -41,9 +35,18 @@ class Address extends React.Component {
         } ${
             this.props.location.streetAddressLine2()
         }`;
+
+        let size;
+
+        try {
+            size = window.innerWidth - 30;
```
DanielHeath: Don't we have a max-width on the container (800px or so IIRC)?

DanielHeath: Also the api won't provide bigger than 640 even if you ask for it.

DanielHeath:
++ other than the image size thing

src/components/StaticMap.js
danni: I'm thinking of screens smaller than 640.

PR #218, opened 2015-10-22
Update bylines from UAT

DanielHeath:


danni:
Etc ends with a full stop. Otherwise ++

DanielHeath:
Not for bylines

danni:
++

PR #219, opened 2015-10-22
Remove transport time (to be re-instated once google integration works)

DanielHeath:


danni:
Why not leave in the directions link? Also why not just make it work. It works perfectly everywhere but Victorian?

DanielHeath:
Good point - https://github.com/ask-izzy/ask-izzy/pull/234 was about as easy as removing it.

PR #224, opened 2015-10-26
[Needs discussion] Ensure page transitions have completed before loading timeshift

DanielHeath:
We've had a bunch of spurious test failures around timeshift not loading correctly. 

I suspect the issue is that the page is transitioning across URLs and losing the script. 
This change seems to resolve it.

### DIFF
--- a/test/steps/datetime.js
+++ b/test/steps/datetime.js
@@ -9,6 +9,8 @@ import Yadda from "yadda";
 import fs from "fs";
 import moment from "moment";
 
+import { documentReady } from "./browser";
+
 import unpromisify from "../support/yadda-promise";
 
 module.exports = (function() {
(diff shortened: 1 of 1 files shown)


DanielHeath:
This doesn't seem to fix it :/

DanielHeath:
Ok, this hasn't fixed it. Next idea: Build timeshift into the webpack bundle (it's pretty small) and make it use `sessionstorage` so we can stub it reliably across page loads.

DanielHeath:
@danni Are you happy for me to go ahead and make time mocks persistent across page loads?

danni:
I'm curious why we're now getting page loads and never used to.

danni:
Have you confirmed something isn't triggering a spurious page reload because it's using an `<a>` instead of a `<Link>` etc?

DanielHeath:
I'm curious also, but investigating it sounds like something that could eat a week of productivity. 

Besides which, making it persistent would remove a trap for any future developer who works with these tests.

danni:
I'd rather fix the actual bug it's showing up, rather than badger it around in a way that sounds later much harder to debug.

PR #227, opened 2015-10-26
Upgrade to new flow

DanielHeath:


.flowconfig
```diff
@@ -16,6 +16,8 @@ module.system=node
 suppress_type=$FlowIssue
 suppress_comment= \\(.\\|\n\\)*\\flow:disable
 strip_root=false
+esproposal.class_static_fields=warn
```
danni: What do these?

DanielHeath: We should be able to get rid of many of the flow:disable comments. Also valid is `=ignore`.

danni: Perhaps it should be ignore?

danni:
++

PR #228, opened 2015-10-26
Fix most of the scrolling bugs when navigating

DanielHeath:


DanielHeath:
It looks like this fixes scroll for `forward` and `back` but not `<Link>`

src/pages/ResultsListPage.js
```diff
@@ -101,8 +101,11 @@ class ResultsListPage extends BaseResultsPage {
     }
 
     renderPersonalisationLink(): ReactElement {
+        const current = this.props.location.pathname;
+        const separator = current.endsWith("/") ? "" : "/";
```
danni, line 105: There's a method for this in Node URL.

DanielHeath: And also in the browser?

DanielHeath: Ah right - Location is a URL object so that should be fine.

DanielHeath, line 105: ergh, relative URLs lose the last segment of the path when using `url.resolve`. Don't think that will work.

danni, line 105: Do we want the last segment then?

DanielHeath, line 105: Yes - we need eg `/category/housing/personalise/summary`, not `/category/personalise/summary`

danni, line 105: Hmm. Could we parse the URL, append to the path array and reoutput?

DanielHeath, line 105: The path isn't an array - we could split it on '/' and remove all the empty strings from the resulting array, but that also sounds like hacks.

danni:
++

PR #230, opened 2015-10-26
Add 'skip' button for personalisation

DanielHeath:


src/pages/personalisation/BaseQuestion.js
```diff
@@ -172,6 +172,11 @@ class BaseQuestion extends React.Component {
     static getSearch(request: iss.searchRequest): ?iss.searchRequest {
         let value = this.answer;
 
+        if (value == "(skipped)") {
```
danni, line 187: Can we do something less string based?

danni:
Needs tests. Also just to confirm this isn't applied to location?

src/pages/personalisation/BaseQuestion.js
DanielHeath, line 187: It's got to round-trip safely through JSON and also render nicely with `toString()` - so it might be difficult.

DanielHeath:
Correct - location isn't a subclass of BaseQuestion / BaseMultiQuestion.

DanielHeath:
Following design meeting, this is being replaced by a 'Next' button with a right-chevron in the app bar.

src/pages/personalisation/Location.js
```diff
@@ -202,6 +199,15 @@ class Location extends React.Component {
             });
     }
 
+    setState(state: Object): void {
```
danni: Not sure about this. Potentially very very surprising?

danni:
++

PR #231, opened 2015-10-26
Move ‘feedback’ link into category list

DanielHeath:


src/components/NavBar.js
```diff
@@ -11,6 +15,9 @@ class NavBar extends React.Component {
     static sampleProps = {default: {}};
 
     render(): ReactElement {
+        const domain = SITE_DOMAIN;
+        const mailLink = `mailto:support@${domain}`;
```
danni: do we need `domain` separate here?

src/components/LinkListItem.js
```diff
@@ -51,31 +50,59 @@ export default class LinkListItem extends React.Component {
             classes.push("has-right-icon");
         }
 
+        if (this.props.to.startsWith('mailto')) {
```
danni: Propose instead we create a second property called href. Which I think `<Link>` already supports?

DanielHeath: Good idea. `<Link>` doesn't support setting `href` instead of `to` but I'll do it at the `LinkListItem` level instead.

DanielHeath:
@danni ++?

src/components/LinkListItem.js
```diff
@@ -51,31 +50,60 @@ export default class LinkListItem extends React.Component {
...
+            // to get an <a> instead of a <Link>
+            return (
+                <a
+                    className={classes.join(" ")}
+                    {...rest}
+                >
+                    {this.renderContent()}
+                </a>
+            );
+        }
```
danni, line 75: `else` for clarity

DanielHeath:
Anything else?

danni:
++

PR #236, opened 2015-10-28
Only show footer on HomePage

DanielHeath:


danni:
Probably can remove `BrandedPage` now? Also confirm this hasn't broken the map view.

DanielHeath:
In the design discussion, talking about the map with @isaacwilliams 

We're going to
- [x] Bump the map min-height up to 1/2 of the screen
- [x] Stop scrolling the map around when changing selection
- [x] Remove the numbers on the markers
- [x] Add a line between results (like the results page)
- [x] Kick user from bookmarked map-page to personalisation if they aren't yet personalised (currently broken)
- [x] Remove street-view map from google map

danni:
Maybe we should keep the numbers when there's > 1 that would both meet QA's expectations and the users? I.e. `(.)` vs `(2)`.

isaacwilliams:
I think we should just get rid of the numbers - they're just confusing and don't really communicate what the intention is

DanielHeath:
@danni ++?

danni:
Can we drop BrandedPage ?

DanielHeath:
@danni Done. Anything else?

danni:
++

PR #237, opened 2015-10-28
Pre-heat flow server during typecheck to stop flow timing out

DanielHeath:


danni:
What happens if it hasn't started by the time you're ready for it?

DanielHeath:
Same thing that happens at the moment (`flow` starts the server if its not running, then waits until its up)

danni:
Does this mean I'll have 2 servers?

DanielHeath:
No - it detects theres an existing one and waits for it. They've done a _really_ nice job of it.

DanielHeath:
I've tested this a bunch of different ways - always appears to work well.

danni:
Nice ++

PR #238, opened 2015-10-28
Revert "Add home link to AppBar"

DanielHeath:
This reverts commit a215ed2b9c6949954080fd4612c4b998b295c90c.


DanielHeath:
After discussing with Isaac, we don't need this now that we're fixing the back button.

danni:
++

PR #240, opened 2015-10-30
Update NodeJS to 4.x

danni:
Fixes memory leak caused by https://github.com/nodejs/node/issues/1309


jscn:
Holy semantic versioning, Batman! That's quite a version jump.

++

PR #242, opened 2015-10-30
Search for services that reduce fees for HCC holders

danni:


src/pages/personalisation/HealthSubcategories.js
```diff
@@ -11,26 +11,28 @@ export default class HealthSubcategories extends BaseMultiQuestion {
             /* eslint-disable max-len */
             "Doctor or nurse": remove("health").append({
                 q: "doctor",
-                // No records are currently marked up as is_bulk_billing
-                // is_bulk_billing: true,
```
mucow, line 15: Not keeping is_bulk_billing for the future implementation anymore? Since you added it in the last time.

danni, line 15: It's not implemented at all. However the HCC holders field means you get cheaper/free service if you're a HCC holder, which is more applicable for most doctors, etc.

mucow:
++

PR #243, opened 2015-11-03
Stop breaking the back button after a reload

DanielHeath:


test/features/10-functional/10-basic/50-history.feature
```diff
@@ -3,10 +3,9 @@ Feature: History navigation
     # As a user
     # When I visit a page
     # I want to be able to use the back button to navigate
-
+    @newSession
```
danni: This is not super BDD.

src/client.js
```diff
@@ -25,12 +25,22 @@ function History() {
...
     function goBack() {
-        if (historyLength > 0) {
-            historyLength--;
-            history.goBack();
-        } else {
-            history.pushState(null, "/");
-        }
+        const currentUrl = location.toString();
+
+        historyLength--;
```
danni: What happens if the history length goes negative? Why even keep the length if we don't check it?

test/features/10-functional/10-basic/50-history.feature
DanielHeath: Would 'Given I am using a new browser' be better?

danni: Yes.

test/steps/browser.js
```diff
@@ -247,11 +249,37 @@ async function assertItemNotChecked(label: string): Promise<void> {
     assert.equal(checked, null);
 }
 
+async function newBrowser(): Promise<void> {
```
danni, line 258: Needs a jsdocstring.

danni:
What happens now with deeplinking?

danni:
I.e. This link here, where should Back go? https://askizzy-staging.infoxchangeapps.net.au/service/1078023-melbourne-youth-support-service-myss-at-frontyard-youth-services

DanielHeath:
@danni I think this is now the best we can do without adding a 'home' button or similar.

danni:
++ looks good

PR #247, opened 2015-11-04
Add referral_info to the eligibility section

danni:


DanielHeath:
Not a huge fan of passing more params than we need via `...object`.

++ anyways as it's an established pattern in Izzy.

PR #250, opened 2015-11-05
Implement google analytics

DanielHeath:


src/client.js
```diff
@@ -64,7 +64,17 @@ function History() {
 }
 
 ReactDOM.render(
-    <Router history={History()}>{routes}</Router>,
+    <Router
+        history={History()}
+        onUpdate={() => {
+            if (typeof window.ga == "function") {
+                ga('send', 'pageview', window.location.toString());
```
danni: Can't this just be a try catch?

src/server/HtmlDocument.js
```diff
@@ -227,6 +230,19 @@ class HtmlDocument extends React.Component {
...
+        <script dangerouslySetInnerHTML={{__html: envConfig}} />
+        { GA_TRACKING_CODE &&
+            <script dangerouslySetInnerHTML={{__html: `
+                window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
+                    ga('create', 'UA-XXXXX-Y', 'auto');
+                    ga('send', 'pageview');
+            `}} />
+        }
+        { GA_TRACKING_CODE &&
+            <script async src='//www.google-analytics.com/analytics.js'></script>
```
danni: This is confusing. Break it all out into a function called `renderAnalyticsBlock()` or something.

src/server/HtmlDocument.js
```diff
@@ -227,6 +230,19 @@ class HtmlDocument extends React.Component {
             name="msapplication-TileImage"
             content="/static/favicons/mstile-144x144.png"
         />
+
+        <script dangerouslySetInnerHTML={{__html: envConfig}} />
+        { GA_TRACKING_CODE &&
+            <script dangerouslySetInnerHTML={{__html: `
+                window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
```
danni: Does React not let you put the contents of a script block conventionally?

src/server/HtmlDocument.js
```diff
@@ -227,6 +230,19 @@ class HtmlDocument extends React.Component {
             name="msapplication-TileImage"
             content="/static/favicons/mstile-144x144.png"
         />
+
+        <script dangerouslySetInnerHTML={{__html: envConfig}} />
+        { GA_TRACKING_CODE &&
```
danni: Moving GA into head will slow down page loads.

src/server/HtmlDocument.js
DanielHeath: The `{}` in the script get interpreted as JSX things

src/server/HtmlDocument.js
DanielHeath: Not on browsers that support `async` which is everything remotely recent (hence google recommending it)

src/server/HtmlDocument.js
danni: Does {{}} work, or is that only for objects? Otherwise a comment saying why.

src/server/HtmlDocument.js
danni: This script block isn't marked async.

DanielHeath: Right, but it's an inline script block - blocks parsing the page until the JS is done, but doesn't require a network request or anything.

DanielHeath: Anyways, that's googles perf recommendation if you're targeting modern browsers.

src/client.js
DanielHeath: It can, but that masks any unrelated errors that occur.

DanielHeath:
@danni Have you got access to update the hosting environment with the GA key?

DanielHeath:
@danni also, ++?

danni:
++

You need to file a systems ticket.

PR #257, opened 2015-11-05
Fix color/size of ResultLisItem heading

DanielHeath:
Per discussion on https://trello.com/c/unNBPW7x/356-correct-colour-of-service-in-service-listing , having implemented this I think we'll need to adjust the size on the rest of the result list if we go ahead with this.

### DIFF
--- a/src/components/HeaderBar.scss
+++ b/src/components/HeaderBar.scss
@@ -1,5 +1,6 @@
 .HeaderBar {
-    /* FIXME check these */
+    @include clearfix;
+
     padding-right: 20px;
     padding-left: 20px;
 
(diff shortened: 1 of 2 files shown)


src/components/HeaderBar.scss
```diff
@@ -1,7 +1,7 @@
 .HeaderBar {
-    /* FIXME check these */
     padding-right: 20px;
     padding-left: 20px;
+    min-height: 110px;
```
danni: Would `@include clearfix;` be better?

DanielHeath: Good idea, I agree.

danni:
++

PR #260, opened 2015-11-09
Add support for accessing dev server from a host other than 'localhost'

DanielHeath:


src/server/render.js
```diff
@@ -22,6 +22,12 @@ export default function render(req, res, next) {
         if (process.env.NODE_ENV === "development") {
             webpackStats = require("./webpack-stats.json");
             delete require.cache[require.resolve("./webpack-stats.json")];
+
+            const replaceLocalhost = (url) =>
+                url.replace('localhost', req.hostname);
```
danni: Hmm. This doesn't look right to me.

danni: Use URL parsing rather than just string replacement.

DanielHeath:
Done.

PR #270, opened 2015-11-11
Use classnames library to construct class strings

DanielHeath:


danni:
++ with great enthusiasm

PR #273, opened 2015-11-11
Clear all returns you to the homepage

DanielHeath:
At the moment it clears your preferences, but leaves you on a page that doesn't really make sense.


danni:
++

PR #275, opened 2015-11-12
Share loaded services between list page and map page

DanielHeath:
I'm pretty sure I can make this cleaner than it is (passing fewer props etc) but this is the general approach I'm planning on.


src/pages/BaseResultsPage.js
```diff
@@ -3,98 +3,20 @@
 import React from "react";
 import { History } from "react-router";
 import reactMixin from "react-mixin";
-import _ from "underscore";
 
 import iss from "../iss";
 import BaseCategoriesPage from "./BaseCategoriesPage";
+import ResultsListPage from "./ResultsListPage";
+import ResultsMapPage from "./ResultsMapPage";
```
danni: Circular import?

danni: Oh I see. Use the router instead to set the appropriate inner component.

danni: Like how the footer works.

DanielHeath: I tried that first, but that instantiates the component & I need to pass props

danni:
Rename the components to ResultsPage, MapComponent ListComponent

src/pages/BaseResultsPage.js
danni: You can still pass props. There's a few places we do things like this. `React.cloneComponentWithProps` will do it for instance.

DanielHeath: Lets not make a ton of GC pressure for the sake of using a library

danni: It's going to be that much GC pressure vs everything else that creates GC pressure. We've never been concerned about this before. It's used approximately everywhere in React already.

DanielHeath: It's used in one place in Izzy, to clone an element with no significant child tree. 
A little different to cloning one of the heaviest elements as part of a page transition.

DanielHeath:
We already have `components/ResultsList.js` which only shows results - this component is really more of a page (header, load more, etc).

DanielHeath:
I've put the 'Home' link icon into this PR as it's otherwise not going to merge at all.

src/components/ResultsMap.js
```diff
@@ -28,8 +31,8 @@ class ResultsMapPage extends BaseResultsPage {
 
     componentDidUpdate(prevProps: Object, prevState: Object) {
         if (this.state.maps &&
-            this.state.objects &&
-            this.state.objects.length) {
+            this.props.objects &&
+            this.props.objects.length) {
```
danni: `_.isEmpty`

src/components/ResultsMap.js
```diff
@@ -106,28 +109,23 @@ class ResultsMapPage extends BaseResultsPage {
         this.setState({selectedServices: services});
     }
 
-    onBackClick(): void {
-        if (_.isEmpty(this.state.selectedServices)) {
-            this.props.history.goBack();
-        } else {
+    onGoBack(event: SyntheticInputEvent): void {
```
danni, line 110: Does Flow know what a SyntheticInputEvent is, vs say Event ?

DanielHeath, line 110: Yes - it knows about the react event system (as a builtin)

src/pages/ResultsListPage.js
```diff
@@ -1,153 +1,66 @@
...
-                this.state.meta && this.state.meta.next ?
-                    <components.ButtonListItem
-                        className="MoreResultsButton"
-                        primaryText="Load more results…"
-                        onTouchTap={this.loadMore.bind(this)}
+            <div className="ResultsListPage">
+                <LoadingResultsHeader {...this.props} />
+                <div className="List results">
+                    {
+                        (this.props.objects && this.props.objects.length) ?
```
danni: `_.isEmpty`

DanielHeath:
Blerg, looks like one of eslints nested dependencies updated and now everything is failing.

src/pages/ResultsPage.js
```diff
@@ -188,6 +146,73 @@ class BaseResultsPage extends BaseCategoriesPage {
...
+        }
+    }
+
+    render(): ReactElement {
+        const Component = this
+            .props
+            .location
+            .pathname
+            .match(/map(\/)?$/) ? ResultsMap
+            : ResultsListPage;
```
danni: We should still make this not-hinge on a regex. It's super fragile, and will surprise anyone who thinks editing the router is enough.

danni: Your other option is to make the list and map views nested in the main page view and then include another level of `<RoutingContext>`. You can then instead of sharing the object in props, share it in `context`.

src/pages/ResultsListPage.js
```diff
@@ -1,153 +1,65 @@
...
-    }
 
-    renderPersonalisationLink(): ReactElement {
-        const current = this.props.location.pathname;
-        const separator = current.endsWith("/") ? "" : "/";
+        /* splice a header before the first crisis service */
+        index = _.findIndex(objects, object => object.crisis);
+        if (index != -1) {
+            /* count hotlines */
+            let nhotlines = _.where(objects, {crisis: true}).length;
```
danni: const ?

src/pages/ResultsListPage.js
```diff
@@ -1,153 +1,65 @@
...
+        } else {
+            objects = [];
         }
-    }
 
-    renderPersonalisationLink(): ReactElement {
-        const current = this.props.location.pathname;
-        const separator = current.endsWith("/") ? "" : "/";
+        /* splice a header before the first crisis service */
+        index = _.findIndex(objects, object => object.crisis);
```
danni, line 23: Is index used anywhere, can we combine this into the `if` ?

src/pages/ResultsPage.js
```diff
@@ -188,6 +146,82 @@ class BaseResultsPage extends BaseCategoriesPage {
...
+    }
+
 }
 
-export default BaseResultsPage;
+export default ResultsPage;
+
+export class ResultsPageListing extends ResultsPage {
+    // flow:disable until flow can track inheritance for ReactComponent
+    component(): ReactComponent {
```
danni, line 217: Could probably just be `static component` ?

danni, line 217: Much better solution though. :)

danni:
Looks pretty good. Generally ++ without further review.

src/pages/ResultsListPage.js
DanielHeath, line 23: It gets used to splice in the crisis line header

src/pages/ResultsPage.js
DanielHeath, line 217: I went with instance in this case since it made the caller read better, but static would also make sense.

src/pages/ResultsListPage.js
DanielHeath: I wonder if there's an eslint setting for 'use const if possible'

PR #277, opened 2015-11-16
Remove coordinating bodies from Izzy search results

danni:


DanielHeath:
If this applies to all categories would it be better in `iss.js`?

danni:
It doesn't, just 99.99% of them.

DanielHeath:
++ then.

PR #282, opened 2015-11-17
Show loading indication when transitioning between services

DanielHeath:


DanielHeath:
Failure is saucelabs <-> travis networking

danni:
++

PR #285, opened 2015-11-17
Add a slide-in animation for the 'next' button

DanielHeath:


danni:
Can we maybe use one of the SASS CSS animation mixins?

DanielHeath:
I generally avoid abstracting things if there's only one instance of it. Are you happy to revisit that if / when we add a second animation?

danni:
I mean one of the existing frameworks. Rather than reinventing a wheel. Especially since other animations, i.e. page transitions will be a thing we want in the future.

danni:
Instead of adding Saffron in our repo why not add a `bower.json` ?

danni:
And then have `bower_components` ignored by Git the same way `node_modules` is.

DanielHeath:
@danni ++?

.travis.yml
```diff
@@ -10,6 +10,7 @@ before_install:
 install:
   - npm config set cache ./node_modules_cache
   - npm install
+  - node_modules/.bin/bower install
```
danni: `npm bin` ?

Dockerfile
```diff
@@ -44,6 +44,8 @@ COPY package.json /app/
 RUN npm install && \
     npm cache clean
 
+RUN node_modules/.bin/bower install
```
danni: Combine with the line above and clean the cache afterwards. Also `npm bin`

bower.json
```diff
@@ -0,0 +1,25 @@
...
+  "private": true,
+  "ignore": [
+    "**/.*",
+    "node_modules",
+    "bower_components",
+    "test",
+    "tests"
+  ],
+  "dependencies": {
+    "saffron": "~0.2.3"
```
danni: Remove `~`. Add this to your typecheck.

package.json
```diff
@@ -11,6 +11,7 @@
     "babel-core": "5.8.25",
     "babel-loader": "5.3.2",
     "babel-runtime": "5.8.25",
+    "bower": "1.6.5",
```
danni: Probably just a dev dep?

.gitignore
```diff
@@ -9,3 +9,4 @@
 /public/static/*.js
 /public/static/*.js.map
 /src/server/webpack-stats.json
+bower_components
```
danni: `/`

danni: Also sorting.

danni:
Otherwise ++

PR #290, opened 2015-11-23
Cache repeated queries to ISS

DanielHeath:


src/iss.js
```diff
@@ -193,10 +194,18 @@ async function attachTransportTimes(
     return services;
 }
 
+let requestObjectsCache = new Cache();
+
 export async function requestObjects(
     path: string,
     data: ?searchRequest,
 ): Promise<searchResults> {
+    let url_ = mungeUrlQuery(path, data);
```
danni: const ?

src/iss.js
```diff
@@ -193,10 +194,18 @@ async function attachTransportTimes(
...
+
+    if (requestObjectsCache.exactHit(url_)) {
+        return requestObjectsCache.exactHit(url_);
```
danni: hit = blah, if hit return hit

src/iss/Cache.js
```diff
@@ -0,0 +1,67 @@
+import url from "url";
```
danni, line 3: Use an existing library.

DanielHeath, line 3: Done. Anything else?

danni:
++

PR #292, opened 2015-11-23
Upgrade eslint & work around upsteam breakage

DanielHeath:


.eslintrc
```diff
@@ -54,6 +54,8 @@
         "no-native-reassign": 2,
         "no-proto": 2,
         "no-return-assign": 2,
+        /* Broken by an update to one of eslints dependencies */
+        "no-redeclare": 0,
```
danni, line 58: Hmm :( Can we force that dep back down?

DanielHeath, line 58: Not without some kind of shrinkwrap (which we used to have but dropped because it generated a lot of spurious diffs).

DanielHeath:
https://github.com/eslint/eslint/issues/4481 was the issue.

PR #296, opened 2015-11-24
Store user preferences in a cookie rather than sessionStorage

DanielHeath:


danni:
Personally I'd use localstorage.

DanielHeath:
Localstorage would also work. As discussed - I can't think of a way that it would be meaningfully different.

DanielHeath:
I've switched to localstorage as recommended. 

Slightly concerned that it shares storage between tabs, but I think that's enough of an edge-case for now.

package.json
```diff
@@ -23,6 +23,7 @@
     "history": "1.13.1",
     "immutable": "3.7.5",
     "js-yaml": "3.4.2",
+    "koekje": "0.0.4",
```
danni: Is this still needed?

src/storage.js
```diff
@@ -3,16 +3,22 @@
...
-
 class Storage {
 
     static getItem(key: string): ?(string|number|boolean) {
-        return sessionstorage.getItem(key);
+        if (typeof localStorage == "undefined") {
+            return null;
+        }
+
+        return localStorage.getItem(key);
```
danni, line 13: Out of interest why is this not super upsetting flowtype?

DanielHeath, line 13: It knows about `localStorage` being a thing (and the rest of the DOM stuff)

package.json
DanielHeath: Good catch, no.

danni:
++

PR #297, opened 2015-11-24
Implement recommendations from react-a11y

DanielHeath:


danni:
++ but check how `:focus` overrides on other browsers.

DanielHeath:
It looks fine in chrome. 
I've raised https://github.com/rackt/react-a11y/issues/104 to see if we can get unified event handling.

danni:
++

PR #308, opened 2015-12-07
Float the done button on bottom of screen on long multi-select lists

isaacwilliams:
The "done" / "none of these" button floats on bottom of screen when the list is longer than the height of the screen

### DIFF
--- a/src/pages/BasePersonalisationPage.scss
+++ b/src/pages/BasePersonalisationPage.scss
@@ -112,4 +112,24 @@
         text-align: center;
     }
 
+    .hasFloatingDoneButton {
+        padding-bottom: 80px;
+
+        .done-button {
+            @include riseIn(1s, 0s, forwards);
+
+            position: fixed;
+            bottom: 0;
+            max-width: $max-width;
+            border-top: 1px solid $brand-bg-light;
+            background: $brand-bg-light;
+
+            > .FlatButton {
+                margin-top: 15px;
+                margin-bottom: 15px;
+            }
+        }
+
+    }
+
 }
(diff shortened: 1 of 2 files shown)


src/pages/BasePersonalisationPage.scss
```diff
@@ -112,4 +112,24 @@
         text-align: center;
     }
 
+    .hasFloatingDoneButton {
+        padding-bottom: 80px;
+
+        .done-button {
+            @include riseIn(1s, 0s, forwards);
+
+            position: fixed;
```
danni, line 121: Flexboxes would be better for this as they will push down as required. Have a look at how you do a sticky footer with flex boxes (already in our code).

src/pages/personalisation/BaseMultiQuestion.js
```diff
@@ -118,11 +121,30 @@ class BaseMultiQuestion extends BaseQuestion {
...
     render(): ReactElement {
         let selected = this.selected;
 
         return (
-            <div className="BaseMultiQuestion">
+            <div
+                className={
+                    classnames(
+                        "BaseMultiQuestion",
+                        this.shouldRenderFloatingDoneButton()
```
danni: `classnames` can take a class name and a truthy statement to evaluate: https://github.com/JedWatson/classnames

src/pages/BasePersonalisationPage.scss
isaacwilliams, line 121: I'm not sure that'll produce the correct behaviour - want to have it float over content, not be pushed down

danni, line 121: Okay, if that's your intentional choice then disregard.

isaacwilliams, line 121: I think it miiight be possible to do what I want using flex box but I can't find a good example

DanielHeath:
Failed PR build is due to master having a bad commit

DanielHeath:
++

PR #309, opened 2015-12-08
Don't ask personalisation questions which have already been answered

DanielHeath:


src/pages/PersonalisationWizardPage.js
```diff
@@ -39,9 +39,10 @@ class PersonalisationWizardPage extends BasePersonalisationPage {
     }
 
     nextSubPage(): ?ReactClass {
-        const nextSubpageIdx = this.currentComponentIdx + 1;
-
-        return this.personalisationComponents[nextSubpageIdx];
+        return this.personalisationComponents.find((component, idx) =>
+            (idx > this.currentComponentIdx) &&
+            (!component.getSearch({}))
+        );
```
danni, line 45: Overly complicated.

Just use `for component of ... { if (foo) return bar; }`

DanielHeath, line 45: That sounds more complicated rather than less

DanielHeath, line 45: `find a thing matching this` vs `loop through the things and return if one matches this`

DanielHeath:
https://trello.com/c/8yFtsizm/256-ui-save-personalisation-details-in-a-cookie-small

src/pages/PersonalisationWizardPage.js
DanielHeath, line 45: As discussed - will not work as we need to check the index (which isn't available in a comprehension).

DanielHeath, line 45: @danni anything else?

PR #319, opened 2015-12-08
Remove low-value flaky test

DanielHeath:
This just keeps failing and isn't particularly relevant.


danni:
Can we put an `@skip` or `@xfail` on it or somesuch?

danni:
Otherwise I'm happy to remove this.

DanielHeath:
I could implement something like that, but it's really not worth it for this test (+ ime things like that just end up bitrotting)

PR #320, opened 2015-12-08
Removed catchment true command

robynk:


DanielHeath:
@robynk I think you meant to remove the line a few further down that says `catchment: true`

robynk:
Your comment seems sensible but Danni told me to remove that line.  I'll ask her why.

robynk:
Note to self - if asked to remove a line number and it doesn't look right, ask a question rather than just do it.

PR #327, opened 2015-12-09
Improve google maps link

DanielHeath:
- Use transit directions if it’s quicker
- Only use “Current Location” if you’ve shared your coordinates with us
- Don’t include level / building names as they can confuse gmaps

### DIFF
--- a/src/components/GoogleMapsLink.js
+++ b/src/components/GoogleMapsLink.js
@@ -20,10 +20,7 @@ class GoogleMapsLink extends React.Component {
         const toAddr = this.props.to;
         const {travelTime} = toAddr;
         const mode = travelTime && travelTime.mode == "TRANSIT" ? "r" : "w";
-        const coords = storage.getJSON("coordinates");
-        const start = encodeURIComponent(
-            coords ? "Current Location" : `${storage.getItem("location")}`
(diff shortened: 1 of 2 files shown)


src/components/GoogleMapsLink.js
```diff
@@ -20,15 +17,30 @@ class GoogleMapsLink extends React.Component {
...
-        let start = encodeURIComponent(this.props.from);
-        let query = encodeURIComponent(
-            `${toAddr.streetAddressLine1()} ${toAddr.streetAddressLine2()}`
+        const toAddr = this.props.to;
+        const {travelTime} = toAddr;
+        const mode = travelTime && travelTime.mode == "TRANSIT" ? "r" : "w";
+        const coords = storage.getJSON("coordinates");
+        const start = encodeURIComponent(
+            coords ? "Current Location" : `${storage.getItem("location")}`
+        );
```
danni: I'd always say "Current Location" for the link. Make it Google's problem.

Use case, I'm currently squatting in Footscray, but my Centrelink card says I live in the outer east, so that's where I need to go to get services. I type "Ringwood" into Izzy as my location, but I still need to get to the service from where I currently am.

PR #329, opened 2015-12-09
Make explicit session vs persistent storage available

DanielHeath:


src/storage/strategies.js
```diff
@@ -0,0 +1,35 @@
...
+    persistentStore = sessionStorage;
+    sessionStore = sessionStorage;
+}
+
+// If we can use localStorage for a persistent store, do so.
+if ((typeof localStorage != "undefined") &&
+    canUseStorageImpl(localStorage)) {
+    // localStorage is available
+    persistentStore = localStorage;
+}
```
danni: Why not just make these singletons? Rather than adding them to window under a different name above?

DanielHeath: None of these are added to window

PR #330, opened 2015-12-09
Add done button to 'change your answers' page

DanielHeath:


DanielHeath:
@isaacwilliams the explanatory copy looked confusing after I added the done button, so I've taken it out. Thoughts?

isaacwilliams:
@DanielHeath I just pushed an update switching the order of the buttons and replacing the text - that looks fine now I think

PR #335, opened 2015-12-12
Added service type to finding work category

robynk:


src/constants/categories.js
```diff
@@ -171,7 +171,8 @@ const categories:Array<Category> = [
         icon: icons.Job,
         search: {
             q: "employment -(coordinating bodies)",
-        },
+            service_type: "employment",
+            },
```
kane-c: Outdent this brace so it lines up with `search` then :+1:

PR #339, opened 2015-12-12
Removed space between fullstop & append

robynk:


src/pages/personalisation/JobSubcategories.js
```diff
@@ -11,7 +11,7 @@ export default class JobSubcategories extends BaseMultiQuestion {
         question: "Where do you want to start?",
         answers: {
             /* eslint-disable max-len */
-            "Employment": remove("employment"). append("job searching"),
+            "Employment": remove("employment").append("job searching"),
```
DanielHeath, line 14: Should this be quoted?

PR #344, opened 2015-12-14
Removed extra "

robynk:


src/pages/personalisation/HousingSubcategories.js
```diff
@@ -17,7 +17,7 @@ export default class HousingSubcategories extends BaseQuestion {
                     "("bond scheme) (housing establishment fund) -(coordinating bodies)"),
```
kane-c: Need to do the same fix here

PR #352, opened 2015-12-18
Remove added

robynk:


kane-c:
The build is failing now because line 19 is too long. We try to keep all lines <= 80 characters long. Split that line into two like this:

```
             "Criminal charges": 
                 append("criminal charge criminal law -police"),
```

kane-c:
:star2:

PR #99, opened 2015-08-29
WIP: Screenshot comments

danni:
Adds screenshots/videos to PRs.Depends on #98.

PR #362, opened 2015-12-21
Capture screenshots in travis on supported platforms

DanielHeath:
We've been having lots of intermittent bugs that are hard to track down. This should help...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,4 +44,4 @@ env:
 
   matrix:
     - SELENIUM_BROWSER="firefox:39:Windows 7" SCREENSHOT_FAILURES=base64
-    - SELENIUM_BROWSER="internet explorer:11.0:Windows 7"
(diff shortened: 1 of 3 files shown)


kane-c:
Are the screenshot files accessible from travis somehow?

How about getting #99 fixed up and merged?

DanielHeath:
This just dumps the base64-encoded screenshot into the build log. Super messy but gets the job done.

DanielHeath:
I'd be ecstatic to get #99 going, but this gets us 70% of the way there for 2 minutes of work.

kane-c:
OK. Do you have an easy-ish way of getting the base64 log into a viewable format?

We did a similar thing here in our GitLab CI logs; Danni wrote a userscript to parse the logs and convert it into an inline image.

kane-c:
:+1:

PR #371, opened 2015-12-23
Added commas at end of lines

robynk:


kane-c:
Some of the unit tests will need updating for this.

PR #372, opened 2015-12-23
Fallback when localStorage is unavailable

DanielHeath:
and start using the type system to control storage behavior


DanielHeath:
@kane-c ++?

kane-c:
lgtm

PR #381, opened 2015-12-29
Changed Legal keywords

robynk:


src/constants/categories.js
```diff
@@ -131,7 +131,8 @@ const categories:Array<Category> = [
         byline: "Police, law & fines",
         icon: icons.Legal,
         search: {
-            q: "law -(coordinating bodies)",
+            q: "legal -(coordinating bodies) -permits -ceremonies" +
```
kane-c: Need a space after `ceremonies` or before `-making` otherwise this will result in `-ceremonies-making`

PR #386, opened 2015-12-29
Added keywords to housing category

robynk:


kane-c:
Need a `+` at the end of the line starting with `q`

PR #395, opened 2015-12-30
Added quotation marks around service type

robynk:


kane-c:
Nearly there. Just need to remove the `"` before `housing`

PR #398, opened 2015-12-31
Fix 'sports medicine' in 'something to do' search

kane-c:


jscn:
lolhappynewyears :+1:

PR #407, opened 2016-01-04
Changes to keywords for all subcategories

robynk:


src/pages/personalisation/AdvocacySubcategories.js
```diff
@@ -10,12 +10,16 @@ export default class AdvocacySubcategories extends BaseMultiQuestion {
...
+                .append("ombudsman complaint"),
             "Help working out a problem with a service":
-                remove("advocacy").append("mediation"),
+                remove("consumer issues discrimination disputes advocacy"),
             "Get advice on your rights":
-                remove("advocacy").append("rights advice"),
-            "Someone to speak for you": append("advocacy"),
+                remove("consumer issues discrimination disputes")
+                .remove("advocacy mediation").append("rights advice"),
+            "Someone to speak for you":
```
kane-c: Remove the extra space on the end of this line

kane-c:
Need to change line 4 to this (remove `append`):

``` js
import { remove } from "./BaseQuestion";
```

robynk:
Need to change line 4 to this (remove append):
import { remove } from "./BaseQuestion";

Why would I remove append?  It doesn't seem obvious to me.

kane-c:
The linter is saying that `append` is no longer used in this file, so it is superfluous. I'll remove it.

robynk:
What about .append("rights advice"),

kane-c:
That's with a dot - it's a separate function. Confusing!

robynk:
Ok.  I get it.  I'll never remember it though.

PR #417, opened 2016-01-05
Include the location in the URL bar

DanielHeath:


kane-c:
Should the suburb/state be slugified or something? They might have spaces in the name.

Also, should states be abbreviated? Would there potentially be a URL like `/in/Denman Prospect-Australian Capital Territory`

DanielHeath:
Yes, we'd end up with urls like that. 

I can't think of a reason it'd be a problem to have long URLs though.

PR #427, opened 2016-01-06
Count XHRs in progress and wait for them between test steps

DanielHeath:
This PR makes the documentation for `documentReady` accurate.


DanielHeath:
@kane-c ++?

DanielHeath:
We've had a few tests fail because they don't wait for stuff to load.

kane-c:
++

We use a similar thing in our Django apps

PR #434, opened 2016-01-06
Remove "public" from answer "public housing"

reducetoflow:
removed "public" from answer "Public Housing", includes remove of public from the dict's value for this key.


DanielHeath:
@reducetoflow has this been run past Robyn / Diana ?

DanielHeath:
I'll investigate the flaky CI failures separately

PR #439, opened 2016-01-07
Add minimum_must_match to search queries

DanielHeath:


DanielHeath:
Not that it'll do much until ISS supports it.

src/iss.js
```diff
@@ -394,6 +396,7 @@ export async function search(
     let request_: searchRequest = {
         q: "",
         type: "service",
+        minimum_must_match: "3>50%",
```
kane-c: I'm not sure if `>` is allowed. Docs suggest `<` or `<-`
https://www.elastic.co/guide/en/elasticsearch/reference/1.4/query-dsl-minimum-should-match.html

DanielHeath:
Oh right - sign is backwards. For `50%` I think `<` and `<-` are equivalent, right?

DanielHeath:
@kane-c ++?

kane-c:
++

PR #444, opened 2016-01-07
Added service type to housing category

robynk:


src/constants/categories.js
```diff
@@ -43,7 +43,8 @@ const categories:Array<Category> = [
         search: {
             q: "housing -(coordinating bodies) -(respite care) " +
                 "-(housing information) -hef",
-        },
+            service_type: "housing",
+            },
```
kane-c: This `{` needs to line up with the start of `personalisation`

PR #445, opened 2016-01-07
Fixed messy lines

robynk:


src/constants/categories.js
```diff
@@ -43,8 +43,8 @@ const categories:Array<Category> = [
         search: {
             q: "housing -(coordinating bodies) -(respite care) " +
                 "-(housing information) -hef",
-            service_type: "housing",
-            },
+                service_type: "housing",
```
kane-c: Almost... service_type was right, just the brace needed to move :)
You can put service_type back where it was

robynk: Does that mean the service types in everyday things & finding work are wrong?  I lined it up with them.

kane-c: It needs to be in line with `q`

PR #464, opened 2016-01-10
increase height of appbar to 44px

isaacwilliams:
and add min-width to back button of 44px


kane-c:
:+1:

DanielHeath:
Can we convert this height to a variable instead?

isaacwilliams:
@DanielHeath done.

PR #467, opened 2016-01-11
Retry a couple of times when failing to see something

DanielHeath:
We’ve had a few failures where something had been added to the DOM but not visible yet.


kane-c:
:+1:

PR #470, opened 2016-01-11
Responsive tweaks

isaacwilliams:
Tweaks to responsive layout. 

Crisis lines are now two-across instead of three. Adjusted the positioning of the Open Times collapser button to put it in line with the directions button (which incidentally fixes the problem with a long "Open now until...")

### DIFF
--- a/src/components/ListItem.scss
+++ b/src/components/ListItem.scss
@@ -9,6 +9,10 @@
         padding-bottom: 10px;
 
         font-size: 18px;
+
+        & > div > .leftIcon {
+            top: 15px;
+        }
     }
 
     &.hero {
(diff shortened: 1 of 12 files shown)


src/pages/BasePersonalisationPage.scss
```diff
@@ -118,6 +128,7 @@
         }
     }
 
+
```
DanielHeath: Extra whitespace?

src/components/ListItem.scss
```diff
@@ -9,6 +9,10 @@
         padding-bottom: 10px;
 
         font-size: 18px;
+
+        & > div > .leftIcon {
```
DanielHeath: Can we use classes?

PR #485, opened 2016-01-12
Added quotation marks to school

robynk:
The search string only works in ISS with the quotation marks.


DanielHeath:
Does that change something in an important way?

DanielHeath:
(genuinely curious - I wouldn't have expected that to matter but I don't know ISS well)

robynk:
Without the quotes you get nothing returned, which is also what AI is returning.  In ISS it is quotation marks.  I wondered if it should be brackets in AI because we seem to use them for phrases rather than "

robynk:
I have no idea why it won't work without the quotes.

PR #496, opened 2016-01-14
Avoid storing anything long-term for DV

DanielHeath:
Until you delete all answers, we’ll put everything into
sessionStorage and clear out localStorage.


kane-c:
lgtm

PR #512, opened 2016-01-18
Add terms of use link to footer

isaacwilliams:
Keeping things simple. Added a terms link next to the About link. Refactored footer lists slightly to be more generic.


kane-c:
++

PR #514, opened 2016-01-18
Upgrade to node 5

DanielHeath:


kane-c:
Looks like the build used Node 4. Need to update `.travis.yml`?

DanielHeath:
Good spot :)

DanielHeath:
ergh, a dependency got yanked so the build failed

kane-c:
++ when build passes

PR #515, opened 2016-01-18
Add 'New service' mailto link on about page

DanielHeath:


src/pages/AboutPage.js
```diff
@@ -45,6 +45,34 @@ class AboutPage extends React.Component {
...
                     </p>
+                    <a
+                        href={
+                            "mailto:database@infoxchange.org" +
+                            "?subject=" +
+                            encodeURIComponent("Ask Izzy add site/service") +
+                            "&body=" +
+                            encodeURIComponent(`Service provider name:
+
+Service name:
```
kane-c: It would be nicer to have this indented and use a function to trim the leading whitespace on each line. If that's something trivial to implement then it would be good to see, but otherwise ++

PR #523, opened 2016-01-19
Handle multiple opening times in the same day

DanielHeath:


src/iss/ServiceOpening.js
```diff
@@ -73,13 +73,15 @@ export default class ServiceOpening {
                 end: close,
             }, openingHours);
 
-            openingTimes[open.diff(now(), "days")] = serviceOpeningHours;
-            closingTimes[close.diff(now(), "days")] = serviceOpeningHours;
+            openingTimes.push(serviceOpeningHours)
```
kane-c: Semicolons

src/components/CollapsedOpeningTimes.js
```diff
@@ -74,7 +87,10 @@ export default class CollapsedOpeningTimes extends React.Component {
         ];
 
         let openingHours = _(this.props.object.openingTimes)
-            .sortBy(record => order.indexOf(record.day));
+            .sortBy(record =>
+                (order.indexOf(record.day) * 24) +
+                (parseInt(record.open))
```
kane-c, line 92: What is `record.open`? Is it like `"01:00:00"`? If so, `parseInt` would always just give `1`

DanielHeath, line 92: `parseInt("12:30:00")` gives `12`

DanielHeath, line 92: The idea is that if you're open more than once in a day, we want the opening times to show the morning entry before the afternoon one..

kane-c, line 92: Huh strange. I tried it in the console and it was always 1 but now that I try more it's different. OK, ignore me.

DanielHeath:
@kane-c ++?

kane-c:
++

PR #536, opened 2016-01-21
Report JS and XHR errors to google analytics

DanielHeath:


kane-c:
++

I didn't know GA had exception logging. I'll need to add that elsewhere :)

DanielHeath:
It doesn't - it just has custom event tracking :p

PR #540, opened 2016-01-22
Age groups fix

kane-c:
The parameter `age_groups` was wrong. That is the name of the property in ISS entities, but in search results, the filter is `age_group`.


mucow:
Looks fine

PR #547, opened 2016-01-24
Revert "Make fiddling with search from the console easier"

jscn:
Reverts ask-izzy/ask-izzy#544

The PMs want to do several releases without this change, so it's easiest to just undo the merge for now.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -8,10 +8,8 @@ import storage from "./storage";
 import routes from "./routes";
 import sendEvent from "./google-tag-manager";
 import searchTest from "./search-test";
-import categories from "./constants/categories";
 
 window.searchTest = searchTest;
-window.categories = categories;
 
 /*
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -8,10 +8,8 @@ import storage from "./storage";
 import routes from "./routes";
 import sendEvent from "./google-tag-manager";
 import searchTest from "./search-test";
-import categories from "./constants/categories";
 
 window.searchTest = searchTest;
-window.categories = categories;
```
mucow, line 14: Was this all there was to revert? It just enabled someone to be able to fiddle with some of the data on the client?

jscn, line 14: Yes, that's the whole changeset.

jscn, line 14: Just doing as asked :p

mucow, line 14: Lol ++

PR #548, opened 2016-01-24
Removed minimum should match code

robynk:
This will be removed from staging only so comparative testing can be done by Diana, Mark and David


jscn:
++ lgtm

PR #550, opened 2016-01-25
Revert "Removed minimum should match code"

jscn:
Reverts ask-izzy/ask-izzy#548

Testing shows that removing the `minimum_should_match` parameter _does not_ improve the crisis line behaviour. Since the quality of the search results is somewhat better _with_ this parameter in place, we'll leave it in.


mucow:
++

PR #563, opened 2016-01-27
limit nodejs to 1.5GB of RAM rather than 4GB

sammcj:
limit nodejs single process to 1.5GB of RAM rather than 4GB.

As per https://github.com/nodejs/node-v0.x-archive/wiki/FAQ#what-is-the-memory-limit-on-a-node-process

### DIFF
--- a/script/prod-server
+++ b/script/prod-server
@@ -6,4 +6,4 @@ PATH=`npm bin`:$PATH
 export PORT=${PORT-8000}
 export NODE_ENV=production
 
-node index
+node --max_old_space_size 1536 index


sammcj:
FYI our gitlab builds succeeded: https://gitlab.office.infoxchange.net.au/samm/ask-izzy/builds/19199

DanielHeath:
From my reading, the default for this setting is already 1.5GB

sammcj:
Can you confirm that somehow because at present it's trying to use more than 2GB.

DanielHeath:
I've added you as a repo owner - the CI doesn't work on forks.

DanielHeath:
@sammcj looks like this has broken deploys :(

sammcj:
How did it pass tests?

PR #566, opened 2016-01-27
Update about us page

DanielHeath:


src/pages/AboutPage.js
```diff
@@ -105,6 +107,12 @@ class AboutPage extends React.Component {
                     <p>
                         News Corp’s mission is to inspire, inform and advocate for a better Australia, which is why we are collaborating with REA Group, Google and Infoxchange on Ask Izzy. Each day we reach 7.1 million Australians through our newspapers and websites. We have an important role in communicating the value and existence of this critical social tool that will benefit the vulnerable in our society.
                     </p>
+
```
jscn, line 110: Github's code highlighting seems to be freaking out about here, but not sure I can see why. Maybe the lines should be wrapped to <= 79 chars?

jscn:
Otherwise, ++

PR #568, opened 2016-01-27
[WIP] Add service form

kane-c:
Content and styles will need revision. Thoughts on this approach? It's not ideal but the form was built with jQuery and Bootstrap in mind.

### DIFF
--- a/src/pages/AddServicePage.js
+++ b/src/pages/AddServicePage.js
@@ -1,0 +1,80 @@
+/* @flow */
+
+import React from "react";
+
+import components from "../components";
+
+declare var ISS_URL: string;
+
+class AddServicePage extends React.Component {
+
+    constructor(props: Object): void {
+        super(props);
+
+        this.state = {
+            isFormDone: false,
+        };
+    }
+
+    componentDidMount(): void {
+        if (typeof window !== "undefined") {
+            window.addEventListener("message", (event) => {
+                let origin = event.origin || event.originalEvente.origin;
+
+                if (origin !== ISS_URL || event.data !== "formSubmitted") {
+                    return;
+                }
+
+                this.setState({
+                    isFormDone: true,
+                });
+            }, false);
+        }
+    }
+
+    render(): ReactElement {
+        let history = this.props.history;
+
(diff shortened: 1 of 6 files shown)


DanielHeath:
I thought we were linking to a totally other app to do this (in django or something)

src/pages/AddServicePage.js
```diff
@@ -0,0 +1,101 @@
...
+declare var ISS_URL: string;
+
+/**
+ * Wraps the http request code in a promise.
+ *
+ * @param {Object} obj - data passed to http.request.
+ *
+ * @returns {Promise<Object>} a promise for the request.
+ */
+function _request(obj: XhrOptions) {
```
DanielHeath: Probably better to use the existing function rather than copying it from iss.js

src/pages/AddServicePage.js
```diff
@@ -0,0 +1,101 @@
...
+    constructor(props): void {
+        super(props);
+
+        this.state = {
+            form: "Loading...",
+        };
+    }
+
+    async componentDidMount(): void {
+        let response = await _request({
```
DanielHeath: Oh, I see - we're ajax-ing the form in.

DanielHeath: `https://iss-staging.infoxchangeapps.net.au/add-service-form?form=ask-izzy` doesn't do anything yet, right?

DanielHeath: Wouldn't an iframe be much easier?

kane-c: Perhaps. That was one of the options we considered, however the problem was that it won't inherit the styles of the page - e.g. custom fonts, etc.

I'll investigate an iframe version a bit more.

DanielHeath: I _really_ don't want to be styling the form part-in-izzy and part-in-iss.

kane-c:
Updated. Now uses an iframe instead of hacky Ajax magic.

The only problem now is that the form just uses default Bootstrap styles.

I've added the ability to pass in a href to your own style sheet (`css` parameter on `add-service-form`, but a new stylesheet is required for that.

DanielHeath:
++

PR #570, opened 2016-01-27
Revert "limit nodejs to 1.5GB of RAM rather than 4GB"

jscn:
Reverts ask-izzy/ask-izzy#563

Appears to be preventing node from starting, with `node: bad option: --max_old_space_size`

See also, perhaps:
https://kibana.private.infoxchange.net.au/#/doc/logstash-*/logstash-2016.01.27/docker?id=AVKB2kpZvXyWs98dfW8V&_g=%28%29

### DIFF
--- a/script/prod-server
+++ b/script/prod-server
@@ -6,4 +6,4 @@ PATH=`npm bin`:$PATH
 export PORT=${PORT-8000}
 export NODE_ENV=production
 
-node --max_old_space_size 1536 index
+node index


sammcj:
@jscn how did this pass tests if it caused the app to fail?

DanielHeath:
The tests don't use `prod-server`

sammcj:
:dizzy_face:

jscn:
Due the way source control and CI is set up, as soon as something gets merged, it gets built and release into test and staging. The CI run is not really a gatekeeper on merging/releasging. In fact, the image we deploy is not even the image which gets the tests run on it AFAIK, since the image which gets deployed is built by gitlab, and the image on which the tests are run is built by Travis.

jscn:
Not that packages on npm would ever change between those builds or anything...

jscn:
image/container/whatever... 12 hours in on today and still going

sammcj:
OK, sounds like this is something that probably needs to be addressed in the near future as it really should be the 'real' artefact that's being tested rather than Travis's one off 'build' of the app and dependencies at the time code was committed.

sammcj:
also, sorry for breaking the build kthnxbye :toilet:

PR #571, opened 2016-01-27
Run Izzy under supervisord

DanielHeath:


kane-c:
lgtm

DanielHeath:
Not required now we're going to be serving statically.

PR #575, opened 2016-01-27
Changed doctor/nurse & hospital keywords

robynk:


kane-c:
++ but I think this is redundant due to #577

PR #576, opened 2016-01-27
Disable npm's progress bar

sammcj:
DO NOT MERGE THIS UNTIL TESTED AND PEER REVIEWED

Disabling npm's progress bar improves npm install times by a factor of 2x.
This is done by adding --no-progress to npm install lines according to their documentation.

https://twitter.com/gavinjoyce/status/691773956144119808?s=17

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,8 +41,8 @@ RUN echo "Disabling non-essential packages" && \
 
 # Install the npm deps
 COPY package.json bower.json /app/
-RUN npm install && \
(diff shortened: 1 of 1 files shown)


kane-c:
I believe this doesn't make a difference in CI.

It would be better to add `progress=false` to `.npmrc`

sammcj:
Cool that's a good idea if that works!

DanielHeath:
https://github.com/ask-izzy/ask-izzy/pull/590 should do it

PR #581, opened 2016-01-28
Remove doctor/nurse as search results aren't good enough

DanielHeath:


src/pages/personalisation/HealthSubcategories.js
```diff
@@ -1,6 +1,6 @@
 /* @flow */
 
-import BaseQuestion, { remove, append } from "./BaseQuestion";
+import BaseQuestion, { remove } from "./BaseQuestion";
```
jscn, line 3: This isn't the same `append` being used on L13?

DanielHeath, line 3: no, thats a method call

jscn, line 3: Cool, tired eyes. Looks good.

jscn:
++, let's get it checked in staging before we tag a release

PR #654, opened 2023-08-03
Add inline caching to dockerfile + general cleanup

callumgare:
Docker now has some really cool inline caching options and by taking advantage of that we can make the build time several times faster.

Old: https://github.com/ask-izzy/ask-izzy/actions/runs/5745613326/job/15573875187
New: https://github.com/ask-izzy/ask-izzy/actions/runs/5745888586/job/15574537134


flara-infoxchange (approved):
Looks good

michaeladams (approved):

PR #662, opened 2023-09-13
Show alerts open at first if all are defaultToOpen

callumgare:
https://infoxchange.atlassian.net/browse/AI-333


flara-infoxchange (approved):
All good

PR #670, opened 2024-04-16
AI 344 call button and drop down

YookesDev:
Modified the contacts component and the contact buttons


src/components/PhoneButton.scss
```diff
@@ -68,22 +65,31 @@
...
+                    fill: #{colours.get-static('colour-crisis-white')};
+                    fill: var(--colour-crisis-white);
                 }
             }
         }
     }
 }
+
+.call-text {
+    font-size: 0;
```
michaeladams: Why is the 0 font size required?  I haven't seen this before?

src/components/ContactMethods.js
```diff
@@ -1,109 +1,85 @@
...
+                analyticsEvent={{
+                    event: "Action Triggered - Other Contact Details",
+                    eventAction: "Show other contact details",
+                    eventLabel: null,
+                }}
+            >
+                {phones().slice(1).map(renderContactMethod)}
+            </Collapser>
+
+            {/* Spacer before Email */}            <Spacer/>
```
michaeladams: Could we have the <Spacer/> on a new line, so its indented correctly?

michaeladams (approved):

PR #672, opened 2024-04-22
AI-325-change-search-insubcategories

Rahma9345:


GeorgeSVargas:
Hi Rahma, This was a good PR but I think it could be useful for you to check the organization's Git Workflow Guide (will DM)

PR #675, opened 2024-05-01
Ai 318 changes to support with everydays tasks subcategory description

Rahma9345:


michaeladams (commented):
this one seems to have a lot more changes in it then just changes required for AI-318?

PR #684, opened 2024-05-27
AI-310 changed icons into mui

Rahma9345:


src/components/Accessibility.js
```diff
@@ -2,10 +2,11 @@
 
 import type {Node as ReactNode} from "React";
 import React from "react";
-import icons from "../icons";
+/*import icons from "../icons";*/
```
callumgare: Generally we try to remove unused code rather than comment it out. Otherwise the code base can quickly fill up with commented out code. If we ever need to use some some old code again we can look in the git history to find it.

src/components/Address.js
```diff
@@ -2,12 +2,12 @@
 import type {Node as ReactNode} from "React";
 import React from "react";
 import classNames from "classnames";
-import Map from "@/src/icons/Map";
-import MapSolid from "@/src/icons/MapSolid";
+/*import Map from "@/src/icons/Map";*/
+/*import MapSolid from "@/src/icons/MapSolid";*/
```
callumgare: Similar here, we want to delete these lines rather than commenting them out.

src/components/Email.js
```diff
@@ -2,10 +2,11 @@
 import type {Node as ReactNode} from "React";
 import React from "react";
 
-import icons from "../icons";
+/*import icons from "../icons";*/
```
callumgare: Same here

src/components/MyListResultItem.js
```diff
@@ -3,8 +3,9 @@
 import type {Node as ReactNode, Element as ReactElement} from "React";
 import React from "react";
 
-import icons from "../icons";
+/*import icons from "../icons";*/
```
callumgare: And here

src/components/OpeningTimes.js
```diff
@@ -6,7 +6,8 @@ import React from "react"
 import ServiceOpening from "../iss/ServiceOpening"
 import ScreenReader from "./ScreenReader"
 
-import icons from "../icons"
+/*import icons from "../icons"*/
```
callumgare: And here

src/components/PhoneButton.js
```diff
@@ -5,12 +5,13 @@ import { titleize } from "underscore.string";
 
 import Link from "./base/Link";
 import type {AnalyticsEvent} from "../google-tag-manager";
-import icons from "../icons";
+/*import icons from "../icons";*/
```
callumgare: And here

src/components/PhoneButton.scss
```diff
@@ -76,16 +76,26 @@
                 border: 0;
             }
 
-            .Icon {
+            /*.Icon {
                 path {
                     fill: #{colours.get-static('colour-crisis-white')};
                     fill: var(--colour-crisis-white);
+                }*/
```
callumgare: And here

src/components/ResultListItem.js
```diff
@@ -3,12 +3,13 @@
 import type {Node as ReactNode, Element as ReactElement} from "React";
 import React from "react";
 
-import icons from "../icons";
+/*import icons from "../icons";*/
```
callumgare: And here

PR #692, opened 2024-06-16
Ai 343 expand all and collapse all features in service page

YookesDev:


test/steps/results.js
```diff
@@ -48,27 +49,31 @@ async function seeTheResultsIn(
     label: string,
     table: Array<Object>,
 ): Promise<void> {
-    const getText = element => element.getText();
     const keyToClass = (key) => key.match(/[(](.*)[)]/)[1];
     const selector = label && `.${label.replace(/ /g, "")}`;
+    console.log(table);
```
trentmarinoIX: Is this still needed?

src/components/OpeningTimesList.js
```diff
@@ -0,0 +1,51 @@
...
+    service : Service,
+    withSpacer?: boolean,
+    withIcon?: boolean,
+};
+
+function formatTime(str: string): string {
+    return moment(str, "HH:mm:ss").format("h:mm A");
+}
+
+function OpeningTimesList({ service, withSpacer = false, withIcon = false}: Props): React.Node {
```
trentmarinoIX: Good idea moving it into a separate component

test/steps/results.js
YookesDev: oops omg I keep forgetting to remove my console.logs

src/components/general/Collapser.js
```diff
@@ -1,55 +1,64 @@
...
+};
 
-// By default, if 'collapseMessage' is not defined, then the component will not
-// display a button to collapse once it has been expanded.
 function Collapser({
     className,
+    header,
     expandMessage,
     collapseMessage,
     initiallyExpanded = false,
```
trentmarinoIX: I know this wasn't added in this PR, but `initiallyExpanded` could be removed. I see it's only used here 
https://github.com/ask-izzy/ask-izzy/blob/f3c60c3f39b16ec931607db540bb6096afff1763/src/components/DebugPersonalisation.js#L64

and here https://github.com/ask-izzy/ask-izzy/blob/f3c60c3f39b16ec931607db540bb6096afff1763/src/components/DebugContainer.js#L12

Because you have the `externalCollapsed` prop the `initiallyExpanded` becomes redundant because `externalCollapsed` will still trigger the useEffect on-load.

``` js
useEffect(() => {
        // Runs this
        if (typeof externalCollapsed === "boolean" && externalCollapsed !== isCollapsed) {
            setIsCollapsed(externalCollapsed);
        }
    }, [externalCollapsed]); // Triggered on load
```

This means the initial value will be set like this and not defined by the prop
`const [isCollapsed, setIsCollapsed] = useState(true);` 

And this part of the debug container will look like this

``` jsx
<DebugContainer
            message="Debug ISS Query"
            externalCollapsed={
                !!Storage.getJSON("issParamsOverride")
            }
        >
```

Here are all the places I've found that use this prop
https://github.com/search?q=repo%3Aask-izzy%2Fask-izzy%20initiallyExpanded&type=code

test/steps/results.js
trentmarinoIX: Nah you're all good, happens to the best of us, I forget them all the time

src/components/general/Collapser.js
YookesDev: Cheers Trent! this completely slipped my mind as I was trying out different things to get this collapser working, removed it now from the code so waiting for that to build

src/components/general/Collapser.js
```diff
@@ -34,7 +32,7 @@ function Collapser({
     externalCollapsed,
     icon,
 }: Props): ReactElement<"div"> {
-    const [isCollapsed, setIsCollapsed] = useState(!initiallyExpanded);
+    const [isCollapsed, setIsCollapsed] = useState(true);
```
trentmarinoIX, line 35: Looks good, you will need to change all places that use `initiallyExpanded` as a prop to `externalCollapsed`

trentmarinoIX (approved):
++ Looks good :)

PR #701, opened 2024-07-01
Ai 535 reverting css change on collapser

YookesDev:


flara-infoxchange (approved):
That looks ok to me, reverting prod changes is a bit sus but 'aight

PR #747, opened 2026-04-27
Fix the use of the proxy for share services endpoint

callumgare:
https://infoxchange.atlassian.net/browse/AI-396


flara-infoxchange (approved):
Looking mighty fine

